PR #6, opened 2015-05-26
Propose new-ish formula style

puneetk:
**Summary** 
- defaults live in defaults.yml
- map jinja overrides by grain + merges pillar:lookup
- split all contextually similar states in their own files

### DIFF
--- a/template/map.jinja
+++ b/template/map.jinja
@@ -1,7 +1,26 @@
-{% set template = salt['grains.filter_by']({
-    'default': {
-        'pkg': 'template',
-        'service': 'template',
-        'config': '/etc/template.conf',
-    },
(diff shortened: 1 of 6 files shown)


template/map.jinja
```diff
@@ -1,7 +1,26 @@
-{% set template = salt['grains.filter_by']({
-    'default': {
-        'pkg': 'template',
-        'service': 'template',
-        'config': '/etc/template.conf',
-    },
-}, merge=salt['pillar.get']('template:lookup')) %}
+# -*- coding: utf-8 -*-
+# vim: ft=jinjia
```
babilen: This won't work.

PR #7, opened 2015-06-01
Improve style and jinja too match salt-formula

aboe76:
* Improved map.jinja with same style as in salt-formula
* renamed state file -> config
* renamed state package -> install

Now it is logical install(ation) -> config(ugration) -> service


gravyboat:
Looks good to me, thanks!

PR #10, opened 2016-05-09
Fix pillar name of template

MhdAli:
It should be `template:lookup`, even in the comment above it's
mentioned the correct pillar name :)

### DIFF
--- a/template/map.jinja
+++ b/template/map.jinja
@@ -26,7 +26,7 @@ that differ from whats in defaults.yaml
 
 {## Merge in template:lookup pillar ##}
 {% set template = salt['pillar.get'](
-        'template',
+        'template:lookup',
         default=default_settings.template,
         merge=True
     )


aboe76:
But it is already merged, see line:
https://github.com/MhdAli/template-formula/blob/master/template/map.jinja#L22

MhdAli:
You're right, but it'll be good practice if the lookup is there.

Thanks.

MhdAli:
if I'm not mistaken, this is meant to be like this because we may need to get pillars other than `lookup` sub-group.

for example if we wanted to get pillars from `template:config` then we need to add another merge, but in the current setup you don't need to do that, because it'll get all sub-groups below `template`.

is that correct?

MhdAli:
@aboe76 waiting for your input.

aboe76:
@MhdAli Why do you want to merge template:lookup twice???

I have already showed you that it is merged in the previous map, and the last map is there to merge the pillar template into the map....

PR #16, opened 2018-12-10
Note about formula versioning

k-hamza:
Introduce the concept of formula versioning already described in [SaltStack documentation](https://docs.saltstack.com/en/latest/topics/development/conventions/formulas.html#versioning)


aboe76:
@k-hamza thanks for this

PR #20, opened 2019-02-10
update formula with map.jinja and style guide references, improve REA…

aboe76:
Update map.jinja with better merging.
Update README for style guide references
Updated VERSION and CHANGELOG file

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,31 @@
 template formula
 ================
 
+0.1.1 (2019-02-10)
+
+- Update map.jinja and README
+- Add tags and update VERSION and CHANGELOG
+
+0.1.0 (2019-02-08)
+
+- Update formula with TOFS pattern
+
+0.0.9 (2018-12-11)
+
+- Add Version File
+
+0.0.8 (2018-03-02)
+
+- Align with Saltstack Official Documenation
+
+0.0.7 (2017-06-18)
+
+- Fix typos and comments
+
+0.0.6 (2015-07-1)
+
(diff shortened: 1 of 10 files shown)


VERSION
```diff
@@ -1,2 +1,2 @@
 # should contain the currently released version of the formula
-1.0.0
+0.1.0
```
myii: `0.1.1` to match the value in [`CHANGELOG.rst`](https://github.com/saltstack-formulas/template-formula/pull/20/files#diff-de0a9b67ffe2874a076b5fdb15bad484R4).

CHANGELOG.rst
```diff
@@ -1,6 +1,31 @@
 template formula
```
myii: Excellent, tagging releases has already started!  May I recommend standardising our changelogs according to the format set out at https://keepachangelog.com/en/1.0.0/?

aboe76: I pushed the tags myself. :)

aboe76: thats on the todo list...I hope....

VERSION
aboe76: yeah but I need to have a commit hash to make a tag.. I think........

CHANGELOG.rst
myii: @aboe76 Yes, I've already pulled and appreciated the tags in `gitk`!  Looks great, thanks for running with that.

VERSION
myii: But when this is merged, it will _be_ `v0.1.1`.

PR #23, opened 2019-02-12
Fix `map.jinja` and add more OSes

myii:
Fixed issues with `map.jinja`.  Since this formula is for reference purposes, I've added the other OS values found in other formulas, in both `osfamilymap.yaml` and `osmap.yaml`.

### DIFF
--- a/template/osfamilymap.yaml
+++ b/template/osfamilymap.yaml
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-                                                                                             
-# # vim: ft=yaml
+# vim: ft=yaml
+#
+# Setup variables using grains['os_family'] based logic.
+# Only add key:values that differ from `defaults.yaml`.
 ---
 Debian:
   pkg: template-debian
@@ -12,6 +15,17 @@ RedHat:
 Suse:
   pkg: template-suse
 
+Gentoo: {}
+
 Arch:
   pkg: template-arch
-  
+
+Alpine: {}
+
+FreeBSD: {}
+
+OpenBSD: {}
+
+Windows: {}
+
+MacOS: {}
--- a/template/osfamilymap.yaml
+++ b/template/osfamilymap.yaml
@@ -2,7 +2,10 @@
 # vim: ft=yaml
 #
 # Setup variables using grains['os_family'] based logic.
-# Only add key:values that differ from `defaults.yaml`.
+# You just need to add the key:values for an `os_family` that differ
+# from `defaults.yaml`.
+# Only add an `os_family` which is/will be supported by the formula
+# (empty `os_family`s do not need to be listed, just added here as an example).
 ---
 Debian:
   pkg: template-debian
@@ -19,6 +22,8 @@ Gentoo: {}
 
 Arch:
   pkg: template-arch
+  service:
+    name: service-arch
 
 Alpine: {}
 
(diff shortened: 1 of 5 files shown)


template/osfamilymap.yaml
```diff
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-                                                                                             
-# # vim: ft=yaml
+# vim: ft=yaml
+#
+# Setup variables using grains['os_family'] based logic.
+# Only add key:values that differ from `defaults.yaml`.
```
javierbertoli: Is it possible that adding all these empty `os_family`s hashes might confuse a newcomer and let her assume that they have to be listed even if not covered by the formula?

In that case, perhaps the line above can be rephrased, like 

```yaml
# You just need to add the key:values for an `os_family` that differ from `defaults.yaml`.
# Only add an `os_family` which is/will be supported by the formula (empty `os_family`s
# do not need to be listed, just added here as an example)
```

template/osfamilymap.yaml
```diff
@@ -12,6 +15,17 @@ RedHat:
 Suse:
   pkg: template-suse
 
+Gentoo: {}
+
 Arch:
   pkg: template-arch
-  
+
```
javierbertoli: Perhaps change one of these to
```yaml
Arch:
  service:
    name: template-arch
```
as an example that different vars can be added for different `os_family`s?

template/osmap.yaml
```diff
@@ -1,8 +1,15 @@
 # -*- coding: utf-8 -*-
 # vim: ft=yaml
+#
+# Setup variables using grains['os'] based logic.
+# Only add key:values that differ from `defaults.yaml` + `os_family.yaml`.
 ---
```
javierbertoli: Same as above

template/osmap.yaml
```diff
@@ -1,8 +1,15 @@
...
+Ubuntu:
+  pkg: template-ubuntu
+
 Fedora:
   pkg: template-fedora
```
javierbertoli: Same as above

template/osfamilymap.yaml
myii: Absolutely, just wanted to revive the useful comment that was in the `map.jinja` previously and show a list of OSes that formulas support.  I do feel like I've come across the use of `{}` before somewhere... perhaps a situation where the OS family (e.g. `Debian`) makes no changes to the defaults yet the OS derivative (e.g. `Ubuntu`) does?  Maybe just used to explicitly define the OSes?

In any case, your suggestion is excellent and I'll incorporate that.

template/osfamilymap.yaml
myii: Yes, definitely.  I was leaving this for a future PR, as we (hopefully) adapt the formula to actually be functional.  But another example right now is good.

template/osmap.yaml
myii: Agree.

template/osmap.yaml
myii: OK.

myii:
@javierbertoli Thanks for the excellent review.  I'll get those done soon and pushed into this PR.

myii:
@javierbertoli I've applied those changes in b5c4987.  Thanks again for your feedback.

myii:
@aboe76 @javierbertoli Thanks to both of you.

PR #25, opened 2019-02-12
 map.jinja: use grains.filter_by instead of defaults.merge

alxwr:
* map.jinja: use grains.filter_by instead of defaults.merge because defaults.merge does not work with salt-ssh. https://github.com/saltstack/salt/issues/51605
* Added osfingermap.yaml.
* Updated changelog and version. (Also set a tag in my repo, not sure how GitHub handles this.)

### DIFF
--- a/template/map.jinja
+++ b/template/map.jinja
@@ -2,21 +2,21 @@
 # vim: ft=jinja
 
 {## Start imports as ##}
-{% import_yaml 'template/defaults.yaml' as defaults %}
+{% import_yaml 'template/defaults.yaml' as default_settings %}
 {% import_yaml 'template/osfamilymap.yaml' as osfamilymap %}
 {% import_yaml 'template/osmap.yaml' as osmap %}
+{% import_yaml 'template/osfingermap.yaml' as osfingermap %}
 
-{## Merge the osfamilymap ##}
-{% set osfamily = salt['grains.filter_by'](osfamilymap, grain='os_family') or {} %}
-{% do salt['defaults.merge'](defaults['template'], osfamily) %}
-
-{## Merge the osmap ##}
-{% set os = salt['grains.filter_by'](osmap, grain='os') or {} %}
-{% do salt['defaults.merge'](defaults['template'], os) %}
-
-{## Merge the lookup ##}
-{% set lookup = salt['pillar.get']('template:lookup', default={}) %}
-{% do salt['defaults.merge'](defaults['template'], lookup) %}
+{% set defaults = salt['grains.filter_by'](default_settings,
+    default='template',
+    merge=salt['grains.filter_by'](osfamilymap, grain='os_family',
+      merge=salt['grains.filter_by'](osmap, grain='os',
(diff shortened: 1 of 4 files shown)


myii:
@alxwr Excellent, looking good at a first glance.  Glad you added `osfingermap` as well.  When I get a bit more time, I'll just run a couple of tests and then we're good to go.  Thanks for the PR!

alxwr:
@myii Thanks for testing!

template/osfingermap.yaml
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# vim: ft=yaml
+#
+# Setup variables using grains['osfinger'] based logic.
+# You just need to add the key:values for an `osfinger` that differ
+# from `defaults.yaml` + `os_family.yaml`.
```
myii: @alxwr If you're still around, then this will be:

```yaml
`defaults.yaml` + `os_family.yaml` + `osmap.yaml`
```

If not, I'll merge anyway and add that to a future PR.

alxwr: Done :-) Thanks for the review!

myii: @alxwr Don't go anywhere!  I've got another point that needs to be resolved.  Give me 2-3 minutes...

template/map.jinja
```diff
@@ -5,18 +5,18 @@
 {% import_yaml 'template/defaults.yaml' as defaults %}
```
myii: `defaults` => `default_settings`!

template/map.jinja
```diff
@@ -5,18 +5,18 @@
...
-{% do salt['defaults.merge'](defaults['template'], osfamily) %}
-
-{## Merge the osmap ##}
-{% set os = salt['grains.filter_by'](osmap, grain='os') or {} %}
-{% do salt['defaults.merge'](defaults['template'], os) %}
-
-{## Merge the lookup ##}
-{% set lookup = salt['pillar.get']('template:lookup', default={}) %}
-{% do salt['defaults.merge'](defaults['template'], lookup) %}
+{% set defaults = salt['grains.filter_by'](default_settings,
```
myii: Because of this change here.

template/map.jinja
alxwr: Sry, I missed that. (It's late here.)

template/map.jinja
alxwr: > @alxwr Just need this last fix and we're good to go, thanks!

Should be fixed now. Thanks for reviewing!

myii:
@alxwr Fantastic, thanks for the extra effort.  Merged.

myii:
@noelmcloughlin has a PR (https://github.com/saltstack-formulas/firewalld-formula/pull/29) that mentions that there is no `osfinger` (and `osmajorrelease`) grain on Arch Linux.  It's worth investigating whether that will result in errors in `map.jinja`, now that it is based on `osfinger`.

PR #28, opened 2019-02-15
Improve TOFS implementation for increased configurability

myii:
* As discussed between comments:
  - https://freenode.logbot.info/saltstack-formulas/20190214#c1995273
  - https://freenode.logbot.info/saltstack-formulas/20190214#c1995487

---

I'll explain this inline and in further comments.

---

- [x] `prefix` removed -- ultimately replaced by `tpldir` to make `files_switch` macro fully-portable.
    - [x] Potential bug: Confirm that this works properly in `macros.jinja` when using the `{% from ... %}` directive from an `.sls` file in a nested directory (e.g. `pkg/install.sls`).
- [x] `path_prefix` configurable.
- [x] `files_dir` configurable.
- [x] `files_switch_list` configurable.
- [x] `default` configurable.
- [x] `files_switch` pillar values extended beyond searching for grains only -- allowing any number of custom paths to be defined (evaluated as literal paths).
- [x] Duplication removed (single `for` loop in `files_switch`).
- [x] Update `pillar.example` -- current scheme (working with existing formulas using TOFS).
- [x] Update `pillar.example` -- proposed scheme (improved layout with minor change required to existing formulas using TOFS).
- [x] Pillar structure resolved (discussion required) -- see [comment](https://github.com/saltstack-formulas/template-formula/pull/28#issuecomment-463911059) below.
    * PR is based on the proposed `pillar.example` scheme.
- [ ] Update all documentation and inline comments as required.
- [x] Check and update tests to ensure they are passing.

### DIFF
--- a/template/config.sls
+++ b/template/config.sls
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 # vim: ft=sls
 
-{% from "template/map.jinja" import template with context %}
-{% from "template/macros.jinja" import files_switch with context %}
+{%- from "template/map.jinja" import template with context %}
+{%- from "template/macros.jinja" import files_switch with context %}
 
 include:
   - template.install
@@ -10,7 +10,13 @@ include:
 template-config:
   file.managed:
     - name: {{ template.config }}
-    - source: {{ files_switch('template', ['example.tmpl']) }}
+    - source: {{ files_switch(
+                    salt['config.get'](
+                        tpldir ~ ':tofs:files:template-config',
+                        ['example.tmpl', 'example.tmpl.jinja']
+                    )
+              ) }}
     - mode: 644
     - user: root
     - group: root
+    - template: jinja
--- a/template/files/default/example.tmpl.jinja
+++ b/template/files/default/example.tmpl.jinja
@@ -3,4 +3,4 @@
 # Your changes will be overwritten.
 ########################################################################
 
-This is an example file from SaltStack template-formula.
+This is another example file from SaltStack template-formula.
--- a/template/macros.jinja
+++ b/template/macros.jinja
@@ -1,8 +1,7 @@
-{%- macro files_switch(prefix,
-                       files,
+{%- macro files_switch(files,
                        default_files_switch=['id', 'os_family'],
                        indent_width=6) %}
-  {#
+  {#-
     Returns a valid value for the "source" parameter of a "file.managed"
     state function. This makes easier the usage of the Template Override and
     Files Switch (TOFS) pattern.
@@ -43,22 +42,42 @@
             - salt://xxx/files/default/etc/yyy/zzz.conf.jinja
           - template: jinja
   #}
-  {%- set path_prefix = prefix|replace(':', '/') %}
-  {%- set files_switch_list = salt['pillar.get'](prefix ~ ':files_switch',
-                                           default_files_switch) %}
-  {%- for grain in files_switch_list if grains.get(grain) is defined %}
-    {%- for file in files %}
-      {%- set url = '- salt://' ~ '/'.join([path_prefix,
-                                            'files',
-                                            grains.get(grain),
-                                            file.lstrip('/')]) %}
+  {#- Get the `topdir` from `tpldir` #}
+  {%- set topdir = tpldir.split('/')[0] %}
+  {%- set path_prefix = salt['config.get'](topdir ~ ':tofs:path_prefix', topdir) %}
+  {%- set files_dir = salt['config.get'](topdir ~ ':tofs:dirs:files', 'files') %}
+  {%- set files_switch_list = salt['config.get'](
+      topdir ~ ':tofs:files_switch',
+      default_files_switch
+  ) %}
+  {#- Only add to [''] when supporting older TOFS implementations #}
+  {%- for path_prefix_ext in [''] %}
+    {%- set path_prefix_inc_ext = path_prefix ~ path_prefix_ext %}
+    {#- For older TOFS implementation, use `files_switch` from the pillar #}
+    {#- Use the default, new method otherwise #}
+    {%- set fsl = salt['pillar.get'](
+        topdir ~ path_prefix_ext|replace('/', ':') ~ ':files_switch',
+        files_switch_list
+    ) %}
+    {#- Append an empty value to evaluate as `default` in the loop below #}
+    {%- if '' not in fsl %}
+      {%- do fsl.append('') %}
+    {%- endif %}
+    {%- for fs in fsl %}
+      {%- for file in files %}
+        {%- if fs %}
+          {%- set fs_dir = salt['config.get'](fs, fs) %}
+        {%- else %}
+          {%- set fs_dir = salt['config.get'](topdir ~ ':tofs:dirs:default', 'default') %}
+        {%- endif %}
+        {%- set url = '- salt://' ~ '/'.join([
+            path_prefix_inc_ext,
+            files_dir,
+            fs_dir,
+            file.lstrip('/')
+        ]) %}
 {{ url | indent(indent_width, true) }}
+      {%- endfor %}
     {%- endfor %}
   {%- endfor %}
-    {%- for file in files %}
-      {%- set url = '- salt://' ~ '/'.join([path_prefix,
-                                            'files/default',
-                                            file.lstrip('/')]) %}
-{{ url | indent(indent_width, true) }}
-    {%- endfor %}
 {%- endmacro %}


template/config.sls
```diff
@@ -1,16 +1,17 @@
...
+{%- from "template/macros.jinja" import files_switch with context %}
 
 include:
   - template.install
 
 template-config:
   file.managed:
     - name: {{ template.config }}
-    - source: {{ files_switch('template', ['example.tmpl'] }}
+    - source: {{ files_switch(['example.tmpl', 'example.tmpl.jinja']) }}
```
myii: No requirement to provide the `prefix` here any more.  Done just once in the `files_switch` macro.

template/config.sls
```diff
@@ -1,16 +1,17 @@
...
     - mode: 644
     - user: root
     - group: root
+    - template: jinja
```
myii: Using a `jinja` template in order to show the `source` of the output file for debugging purposes.

template/files/default/example.tmpl
```diff
@@ -1,3 +1,6 @@
-# Managed by saltstack
+########################################################################
+# File managed by Salt at <{{ source }}>.
```
myii: This will show the `source` of the file in the header of the managed file.

template/files/default/example.tmpl.jinja
```diff
@@ -0,0 +1,6 @@
+########################################################################
+# File managed by Salt at <{{ source }}>.
```
myii: Adding this file for further debugging.  Using `source` again as before.

template/macros.jinja
```diff
@@ -1,8 +1,7 @@
-{%- macro files_switch(prefix,
```
myii: Removed the need for `prefix`.

template/macros.jinja
```diff
@@ -1,8 +1,7 @@
...
-                       files,
+{%- macro files_switch(files,
                        default_files_switch=['id', 'os_family'],
                        indent_width=6) %}
-  {#
+  {#-
```
myii: Without the `-`, it leaves an unnecessary blank line in the debug output:

```salt
    - source:

      - salt://template/files/ABC/example.tmpl
```

template/macros.jinja
```diff
@@ -43,22 +42,30 @@
             - salt://xxx/files/default/etc/yyy/zzz.conf.jinja
           - template: jinja
   #}
-  {%- set path_prefix = prefix|replace(':', '/') %}
-  {%- set files_switch_list = salt['pillar.get'](prefix ~ ':files_switch',
-                                           default_files_switch) %}
+  {%- set formula = 'template' %}
```
myii: Define the `formula` name once and use it multiple times below.

template/macros.jinja
```diff
@@ -43,22 +42,30 @@
...
+  {%- set path_prefix = salt['config.get'](formula ~ ':tofs:path_prefix', formula) %}
```
myii: Using `config.get` to make the `path_prefix` configurable or use `formula` name as the default.

template/macros.jinja
```diff
@@ -43,22 +42,30 @@
...
+  {%- set files_switch_list = salt['config.get'](
```
myii: Now with `config.get` instead of `pillar.get`.

template/macros.jinja
```diff
@@ -43,22 +42,30 @@
...
+  ) %}
   {%- for grain in files_switch_list if grains.get(grain) is defined %}
     {%- for file in files %}
-      {%- set url = '- salt://' ~ '/'.join([path_prefix,
-                                            'files',
-                                            grains.get(grain),
-                                            file.lstrip('/')]) %}
+      {%- set url = '- salt://' ~ '/'.join([
+          path_prefix,
+          salt['config.get'](formula ~ ':tofs:dirs:files', 'files'),
```
myii: Using `config.get` to make the `'files'` directory configurable or use `'files'` by default.

template/macros.jinja
```diff
@@ -43,22 +42,30 @@
...
   {%- endfor %}
-    {%- for file in files %}
-      {%- set url = '- salt://' ~ '/'.join([path_prefix,
-                                            'files/default',
-                                            file.lstrip('/')]) %}
+  {%- for file in files %}
+    {%- set url = '- salt://' ~ '/'.join([
+        path_prefix,
+        salt['config.get'](formula ~ ':tofs:dirs:files', 'files'),
+        salt['config.get'](formula ~ ':tofs:dirs:default', 'default'),
```
myii: Using `config.get` to make the `'default'` directory configurable or use `'default'` by default.

myii:
So by default, with no pillar, etc. set:

```salt
template-config:
  file.managed:
    - name: /etc/template.d/custom-ubuntu.conf
    - source: 
      - salt://template/files/ABC/example.tmpl
      - salt://template/files/Debian/example.tmpl
      - salt://template/files/default/example.tmpl
    - mode: 644
    - user: root
    - group: root
    - template: jinja
```

---

With the following pillar set:

```yaml
template:
  tofs:
    path_prefix: template_alt
    dirs:
      files: files_alt
      default: default_alt
  files_switch:
    - id
    - osfinger
    - os
    - os_family
```

We get:

```salt
template-config:
  file.managed:
    - name: /etc/template.d/custom-ubuntu.conf
    - source:
      - salt://template_alt/files_alt/ABC/example.tmpl
      - salt://template_alt/files_alt/ABC/example.tmpl.jinja
      - salt://template_alt/files_alt/Ubuntu-16.04/example.tmpl
      - salt://template_alt/files_alt/Ubuntu-16.04/example.tmpl.jinja
      - salt://template_alt/files_alt/Ubuntu/example.tmpl
      - salt://template_alt/files_alt/Ubuntu/example.tmpl.jinja
      - salt://template_alt/files_alt/Debian/example.tmpl
      - salt://template_alt/files_alt/Debian/example.tmpl.jinja
      - salt://template_alt/files_alt/default_alt/example.tmpl
      - salt://template_alt/files_alt/default_alt/example.tmpl.jinja
    - mode: 644
    - user: root
    - group: root
    - template: jinja
```

myii:
Looking at the pillar:

```yaml
template:
  tofs:
    path_prefix: template_alt
    dirs:
      files: files_alt
      default: default_alt
  files_switch:
    - id
    - osfinger
    - os
    - os_family
```

May be advantageous to merge these under one key.  Maybe something like:

```yaml
template:
  tofs:
    files_switch:
      - id
      - osfinger
      - os
      - os_family
    path_prefix: template_alt
    dirs:
      files: files_alt
      default: default_alt
```

template/macros.jinja
myii: Isn't there some clever way to get this automatically?  Something built in?

myii:
Haven't done this yet.  The final (really) crazy suggestion but for ultimate configurability:

```salt
    - source: {{ files_switch(['example.tmpl', 'example.tmpl.jinja']) }}
```

* Make the file list configurable via. `config.get` as well(!)

Meaning something like:

```salt
    - source: {{ files_switch(
                    salt['config.get'](
                        formula ~ ':tofs:files:template-config',
                        ['example.tmpl', 'example.tmpl.jinja']
                    )
              ) }}
```

* So that's `<formula>:tofs:files:<state-id>`.
* We'd need to define `formula` at a higher level so that it could be used both here and in the `files_switch` macro.

With an example pillar:

```yaml
template:
  tofs:
    ...
    files:
      template-config:
        - 'example_alt.tmpl'
        - 'example_alt.tmpl.jinja'
```

---

Actually, just tried it and it worked.  So I'll push it here as a second commit so that you can try it for yourselves.

myii:
So with the second commit and the additions to the pillar as mentioned in the previous comment, we get:

```salt
    - source:
      - salt://template_alt/files_alt/ABC/example_alt.tmpl
      - salt://template_alt/files_alt/ABC/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/Ubuntu-16.04/example_alt.tmpl
      - salt://template_alt/files_alt/Ubuntu-16.04/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/Ubuntu/example_alt.tmpl
      - salt://template_alt/files_alt/Ubuntu/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/Debian/example_alt.tmpl
      - salt://template_alt/files_alt/Debian/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/default_alt/example_alt.tmpl
      - salt://template_alt/files_alt/default_alt/example_alt.tmpl.jinja
    - mode: 644
    - user: root
    - group: root
    - template: jinja
```

template/config.sls
```diff
@@ -3,14 +3,20 @@
 
 {%- from "template/map.jinja" import template with context %}
 {%- from "template/macros.jinja" import files_switch with context %}
+{%- set formula = 'template' %}
```
myii: This is ugly but just for testing in this WIP.

myii:
Obviously, this is only a WIP.  If going forwards with any of the changes, the documentation and `pillar.example` would also need to be updated.

template/files/default/example.tmpl
myii: This header is used in other formulas such as the [`nfs-formula`](https://github.com/saltstack-formulas/nfs-formula/blob/095db7256fd0ba8ebd5ff5e6ddf01ceeb15560ad/nfs/files/exports).

It's useful enough in normal cases, just to know where the source template resides.  In the TOFS system, I believe this is indispensable, since there are so many potential templates a file could have used.

template/macros.jinja
daks: I find the `'- salt://'` weird. I would have done `salt://` and adding the dash in the loop, to have the 'url' really be an url, not an item in a YAML list

daks: maybe the same for `salt://' ~ '/'`, why not use `salt:///` directly?

daks:
> May be advantageous to merge these under one key.

I prefer the merging under the `tofs` key.

template/macros.jinja
myii: That particular section was just direct refactoring, didn't look at it in any detail.

So looking now, this is being used to populate the list under the `source` of `file.managed`, thus requiring the dash `-` prefixed.

myii:
> > May be advantageous to merge these under one key.
> 
> I think merging all keys under the `tofs` key makes more sense

@daks I'm being cautious here because the TOFS pattern is already being used in other formulas.  So this change will cause blowback there.  Leaving it open for discussion for now even though I'd prefer if it could be merged going forward.

myii:
@baby-gnu Based on [your comment](https://freenode.logbot.info/saltstack-formulas/20190215#c1997037), added commit dfbdb3e, which has allowed the `files_switch` macro to be simplified to use just the one loop rather than two.  Removes more duplication.

template/macros.jinja
```diff
@@ -43,22 +42,29 @@
...
+  {#- Append an empty value to evaluate as `default` in the loop below #}
+  {%- do files_switch_list.append('') %}
+  {%- for fs in files_switch_list %}
     {%- for file in files %}
-      {%- set url = '- salt://' ~ '/'.join([path_prefix,
-                                            'files',
-                                            grains.get(grain),
-                                            file.lstrip('/')]) %}
+      {%- if fs %}
+        {%- set fs_dir = salt['config.get'](fs, fs) %}
```
myii: If the "grain" isn't found, it uses it as a literal value in the path instead.  This has two noticeable benefits:

1. Allows for additional path matching from the pillar.
1. Easier to debug when not working, since the unmatched value will be shown directly in the URL.

template/config.sls
myii: Just remembered: `tpldir`.

template/macros.jinja
myii: Just remembered: `tpldir`.

myii:
@baby-gnu Using `tpldir` makes the `files_switch` macro fully-portable -- drop-in to any formula without any modifications required.  So the last remaining task is to decide about the pillar layout as mentioned above.  While it will be a hassle in the existing formulas (`zabbix`, `systemd`, etc.), going forward I believe it is more appropriate to merge it all under one key, as mentioned in [my comment above](https://github.com/saltstack-formulas/template-formula/pull/28#issuecomment-463911059) -- @daks concurred with that suggestion.

myii:
Example of dfbdb3e in action, in regards to using any custom path (`any/path/can/be/used/here` in the example below).  With the pillar:

```yaml
template:
  lookup:
    master: template-master
  tofs:
    ...
  files_switch:
    - any/path/can/be/used/here
    - id
    - osfinger
    - os
    - os_family
```

Results in:

```salt
template-config:
  file.managed:
    - name: /etc/template.d/custom-ubuntu.conf
    - source:
      - salt://template/files/any/path/can/be/used/here/example.tmpl
      - salt://template/files/any/path/can/be/used/here/example.tmpl.jinja
      - salt://template/files/ABC/example.tmpl
      - salt://template/files/Ubuntu-16.04/example.tmpl
      - salt://template/files/Ubuntu-16.04/example.tmpl.jinja
      - salt://template/files/Ubuntu/example.tmpl
      - salt://template/files/Ubuntu/example.tmpl.jinja
      - salt://template/files/Debian/example.tmpl
      - salt://template/files/default/example.tmpl
    - mode: 644
    - user: root
    - group: root
    - template: jinja
```

myii:
Pushed two more commits re: `pillar.example`.

1. The first commit adds the whole `tofs` key alongside `files_switch` from before.
1. The second commits shows the proposal of `files_switch` being demoted into the `tofs` key, including the minor change to `macros.jinja`.

template/files/default/example.tmpl
aboe76: nice

myii:
Added 7ea0f5a to fix the bug reported in the discussion [from here](https://freenode.logbot.info/saltstack-formulas/20190215#c1998586).

For each additional state in the `sls`, additional `default` lines are rendered:

```sls
template-config:
  file.managed:
    ...
    - source:
      - salt://template/files/ABC/example.tmpl
      - salt://template/files/Debian/example.tmpl
      - salt://template/files/default/example.tmpl
    ...

formula.autofs.config.master:
  file.managed:
    ...
    - source:
      - salt://template/files/ABC/auto.master
      - salt://template/files/Debian/auto.master
      - salt://template/files/default/auto.master
      - salt://template/files/default/auto.master
    ...

formula.autofs.config.home:
  file.managed:
    ...
    - source:
      - salt://template/files/ABC/auto.homeLdap
      - salt://template/files/Debian/auto.homeLdap
      - salt://template/files/default/auto.homeLdap
      - salt://template/files/default/auto.homeLdap
      - salt://template/files/default/auto.homeLdap
    ...

formula.autofs.config.away:
  file.managed:
    ...
    - source:
      - salt://template/files/ABC/auto.awayLdap
      - salt://template/files/Debian/auto.awayLdap
      - salt://template/files/default/auto.awayLdap
      - salt://template/files/default/auto.awayLdap
      - salt://template/files/default/auto.awayLdap
      - salt://template/files/default/auto.awayLdap
    ...
```

aboe76:
I think this needs a rebase..

myii:
@aboe76 Rebasing is not a problem.  We still need to finalise what we want to do about the remaining tasks, particularly:

> - [ ] Pillar structure resolved (discussion required) -- see [comment](https://github.com/saltstack-formulas/template-formula/pull/28#issuecomment-463911059) below.
>     * PR is based on the proposed `pillar.example` scheme.

@baby-gnu What are your thoughts about this?

myii:
OK, consensus on using `template:tofs:files_switch` instead of `template:files_switch`.  Some refs:

* https://github.com/saltstack-formulas/template-formula/pull/28#issuecomment-463953567
* https://freenode.logbot.info/saltstack-formulas/20190215#c1997883
* https://freenode.logbot.info/saltstack-formulas/20190218#c2006541
* https://freenode.logbot.info/saltstack-formulas/20190218#c2006583

template/macros.jinja
aboe76: @myii, I'm looking into this to incorporate at the systemd-formula,
but I use it here to distinguish multiple files:
systemd:networkd (has tofs)
systemd:timesyncd (has tofs) 
but both should have a different `prefix`
otherwise the files will be in the same place....

myii:
After resolving https://github.com/saltstack-formulas/systemd-formula/pull/17, transferred the changes back to this PR and now ready to go.

There is one stage left that can be covered by a future PR:

> - [ ] Update all documentation and inline comments as required.

aboe76:
merged it!

myii:
@aboe76 Thanks!

PR #34, opened 2019-02-18
Add a testing scaffold with `kitchen-ci`, `inspec` and `travis` support

javierbertoli:
The template can be tested, it's using `pkg: bash` and `service:name: systemd-udevd` in `pillar.example`


aboe76:
@javierbertoli and @myii merged it, Looks good.

PR #43, opened 2019-02-23
feat(toc): use `markdown-toc` directly to update inline

myii:
* https://github.com/jonschlinkert/markdown-toc
  - Generate a markdown TOC (table of contents) for any markdown files.

---

Our Markdown files are quite lengthy, with content still to be added.  Having a TOC is invaluable but manual maintenance is a pain.  Using the available package linked above, the process of maintaining the TOC across all relevant files is automated during the Travis `semantic-release` run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,12 @@ jobs:
         - go get github.com/myii/maintainer
         - maintainer contributor
 
+        # Update Tables of Content in the relevant `.md` files
+        - npm install markdown-toc -D
+        - markdown-toc -i CONTRIBUTING.md
+        # - markdown-toc -i README.md
+        - markdown-toc -i TOFS_pattern.md
+
         # Install all dependencies required for `semantic-release`
         - npm install @semantic-release/changelog@3 -D
         - npm install @semantic-release/exec@3 -D
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ type(scope): subject
 Besides the version bump, the changelog and release notes are formatted accordingly.
 So based on the example above:
 
-> ### Documentation
+> <h3>Documentation</h3>
 > 
 > * **contributing:** add commit message formatting instructions
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,8 @@
 # How to contribute
 
+This document will eventually outline all aspects of guidance to make your contributing experience a fruitful and enjoyable one.
+What it already contains is information about _commit message formatting_ and how that directly affects the numerous automated processes that are used for this repo.
+
 <table><tr><th>Table of Contents</th></tr><tr><td>
 <!-- toc -->
 <!-- tocstop -->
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -45,6 +45,12 @@ jobs:
         - go get github.com/myii/maintainer
         - maintainer contributor
 
+        # Update Tables of Content in the relevant `.md` files
+        - npm install markdown-toc -D
+        - markdown-toc -i CONTRIBUTING.md
+        # - markdown-toc -i README.md
```
myii: I've put this in here since using `README.md` is common across our repos.  That also begs the question -- should we modify the `.rst` to `.md` in this repo, too?

CONTRIBUTING.md
```diff
@@ -1,10 +1,18 @@
 # How to contribute
 
+This document will eventually outline all aspects of guidance to make your contributing experience a fruitful and enjoyable one.
+What it already contains is information about _commit message formatting_ and how that directly affects the numerous automated processes that are used for this repo.
+
+<table><tr><th>Table of Contents</th></tr><tr><td>
+<!-- toc -->
+<!-- tocstop -->
+</td></tr></table>
+
```
myii: I toyed with keeping this simpler (without the table) but found that it is much nicer with it.  This would be a standard copy-paste block that will then be filled automatically by `markdown-toc` during the release phase.

CONTRIBUTING.md
```diff
@@ -19,7 +27,7 @@ type(scope): subject
 Besides the version bump, the changelog and release notes are formatted accordingly.
 So based on the example above:
 
-> ### Documentation
+> <h3>Documentation</h3>
 >
```
myii: This is the workaround when quoting text with headers.  Really, `markdown-toc` should handle this but it would mean maintaining our own fork.

myii:
Thanks for taking the time to review, @javierbertoli.

.travis.yml
javierbertoli: If you make me answer quickly, I'd say that `md` is more common in github projects, but sincerely don't know. I'm OK with both of them (use rst mostly when I have to write pdfs parsing them with [rst2pdf](https://github.com/rst2pdf/rst2pdf)), but I'm no strongly biased for one over the other.

If you know there's a recommended one for github projects, I'd say let's use that one?

myii: I'm not endorsing any particular format, I have to accommodate for three (I prefer `asciidoctor` myself).  Looking at the wider picture, Markdown is the most common, making the barrier of entry lower (i.e. for contributions).  The solutions we're currently using for automation usually work best for, or are limited to, Markdown files.  Then it simply becomes an issue of _consistency_ of formats across the repo.

noelmcloughlin: There is comparison of RST and MD here: http://hyperpolyglot.org/lightweight-markup 
I have used both, have preferred rst when I want to include media/links in the markup.

myii: @noelmcloughlin Thanks for that.  None of us is doubting which is the _superior_ format.  This is really about which is the most _accessible_, both for other contributors as well as the automated solutions that we're introducing.  From that angle, Markdown can't be denied.  We're using it throughout GitHub and everywhere else, wherever heavy-lifting isn't required.  And the docs stored in the main repo are hardly going to be that.  If we ever get around to producing some serious documentation or even populating the formulas' wikis, the other formats come into their own.

PR #45, opened 2019-02-24
ci(kitchen): use `salt-minion` version of `opensuse` to ensure tests run

myii:
* The actual `opensuse/leap` image was passing but not running any
  of the `inspec` tests -- it links to version `15.x`
* `salt-minion` is `42.3`


aboe76:
done

myii:
Also worth noting that the actual `salt-minon` version pre-installed is `2018.3`.

Issue #24, opened 2019-02-12
map.jinja and yaml speed improvements on large formula's


aboe76:
see https://github.com/saltstack/salt/issues/39017

if we can convert some formula's which have a big or multiple yaml files 
we can convert these to json and import those.

oneliner to convert
``` 
python -c 'import sys, yaml, json; json.dump(yaml.load(sys.stdin), sys.stdout, indent=4)' < file.yaml > file.json
```

We could keep the yaml files for easier maintenance but have a nice readme about why
this formula needs json files for importing...

myii:
@aboe76 Sounds good!  Thanks for adding that to the plan.

---

BTW, there's been a discussion on Slack about using the `PillarStack_SDB` Module **alongside** TOFS for even greater efficiency.  I remember you mentioning that you use (or have used) PillarStack.  I'm going to add this to the plan as something to be evaluated.  Dominik Bessler has offered to give us some feedback about this combination as well.  Any thoughts from your experience?

daks:
So the idea is to use JSON instead of YAML because parsing is faster?

I'm not completely ok with keeping both JSON and YAML files with the same information inside, it will just burden the repository and introduce confusion. If we just want to ease maintenance, maybe add doc about converting JSON to YAML, edit YAML, and then convert YAML to JSON.

myii:
@daks Thanks for sharing your views.  Always helps when others share their expertise.

Just to be clear, @aboe76 is discussing this in relation to formulas which contain an _excessive amount_ of YAML.  It may be worth exploring the performance gains from using JSON in those cases.  Just sharing ideas for the time being as we explore how SaltStack Formulas can be improved.

I agree with you about avoiding duplication at all costs.  I'm sure we can find a solution for this if we decide that we would like to go forward with it.

noelmcloughlin:
Many programmers consider YAML controversial and prefer JSON. This is another consideration.

myii:
@noelmcloughlin added a decent [comment on a PR](https://github.com/saltstack-formulas/template-formula/pull/48#discussion_r259827036) that needs to be included here:

> One more thought (per the vision stuff, not this PR) is that salt can support both yaml and json configuration.  All formulas use `yaml/` today but anticipating `json/`  or other data serialisation technology language in the design may make sense - something like this?
> ```
>    template-formula/
>         template/
>           .. etc ..
>              yaml/
>                   defaults.yaml
>                   osmap.yaml
>                   osfamilymap.yaml
[...]

Issue #21, opened 2019-02-11
Establish `template-formula` as the go-to resource for all formula development


myii:
This issue is a collection of ideas taken from the [log of discussions](https://github.com/saltstack-formulas/template-formula/wiki/%5B2019-02-08%5D-Log:-Initial-improvements'-discussions) captured in the wiki.

- [x] Update to new `map.jinja` style -- #20.
- [x] Switch to the TOFS pattern (`files_switch`) -- #19.
- [ ] Add documentation: guides, proposals, etc. -- use the `template-formula` wiki the central location for planning/organisation/reference.
- [ ] Add a test scaffold with `kitchen-ci` and `inspec` (or `testinfra`) examples for the basic tasks (add/remove package, check service, etc).
- [ ] Add the remove state advocated by @noelmcloughlin.
- [ ] Make the formula "actually functional" so that all parts can be run, to aid understanding and quality assurance.
- [ ] Investigate how the TOFS pattern can be used to mitigate the use of `formula:ng` states — most of those can be integrated and removed and only keep one formula.
- [ ] Make this `template-formula` a sticky formula for the whole SaltStack Formulas project.
- [ ] Adapt `template-formula` to not be too focussed on `pkg` -- must also consider tarball releases, github repos, etc.
- [ ] Add a `macros.jinja` template with common macros that "Any formulae would find useful" -- repeatable patterns deserve repeatable code.
- [ ] Ensure that each file does only one thing and do it well (instead of monolithic files), e.g. `pkg/install.sls`, `pkg/clean.sls`, `repo/install.sls`, `repo/clean.sls`, etc.
- [ ] Evaluate #18 to see if it can be merged into this issue.

daks:
Just an idea: document how to create a formula from this template one, inside this repository or/and in saltstack docs

myii:
@daks I've added that to the task list above -- thanks for that suggestion.

alxwr:
@myii I just wanted to say „Thank you!“ :-)
This formula enabled me to implement https://github.com/saltstack-formulas/prometheus-formula _really_ fast. I found the structure and especially your turnkey Travis/kitchen configuration_very_ helpful.

myii:
@alxwr That's really good to hear and it makes all of this effort worthwhile.  There were lots of contributions involved, including your own!  Good to see that you've got `semantic-release` running as well.

[...]

PR #48, opened 2019-02-24
feat(pkg): separate components, use requisites and add `clean` states

myii:
* `tpldot` is the equivalent of `tpldir` but using `.` instead of `/`:
  - https://github.com/saltstack/salt/blob/0c78d7dc894058988d171a28a11bd4a9dbf60266/salt/utils/jinja.py#L120
  - https://github.com/saltstack/salt/blob/0c78d7dc894058988d171a28a11bd4a9dbf60266/salt/utils/templates.py#L145
* Implementation based on upstream issue:
  - https://github.com/saltstack/salt/issues/10838#issuecomment-391718086

---

@noelmcloughlin Requesting your review because this is the implementation of the ideas [you suggested](https://github.com/saltstack-formulas/template-formula/wiki/%5B2019-02-08%5D-Log:-Initial-improvements'-discussions) in our initial discussions (see `22:34`).  Thanks for those ideas.


template/pkg/config.sls
```diff
@@ -20,3 +20,5 @@ template-config:
     - user: root
     - group: root
     - template: jinja
+    - require:
+      - sls: {{ tpldot }}.install
```
myii: It's a shame that relative includes (i.e. `.install`) can't be used when settings a requisite based on an `sls` -- this is an [upstream bug](https://github.com/saltstack/salt/issues/10838) that still hasn't been resolved.

aboe76: I didn't even know they existed, where did you find these...

myii: @aboe76 I've linked to the code in the [first comment](https://github.com/saltstack-formulas/template-formula/pull/48#issue-255703627).

myii:
@aboe76 Thanks for the review.  What would be the right way to do testing for the `clean` states?  Can it be managed in the same run as the existing tests?

template/pkg/clean.sls
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
+
+{%- from salt['file.dirname'](tpldir) ~ "/map.jinja" import template with context %}
+
+template-service-dead:
+  service.dead:
+    - name: {{ template.service.name }}
+    - enable: True
```
javierbertoli: Shouldn't this be ```enable: False```?

template/pkg/config.sls
```diff
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 # vim: ft=sls
 
-{%- from tpldir ~ "/map.jinja" import template with context %}
-{%- from tpldir ~ "/macros.jinja" import files_switch with context %}
+{%- from salt['file.dirname'](tpldir) ~ "/map.jinja" import template with context %}
```
javierbertoli: [Documentation shows usage of this format using a `+` instead of `~`](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.file.html#salt.modules.file.dirname). Even if both formats are correct, shouldn't we try to be consistent with the documented format?

template/pkg/init.sls
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
+
+include:
+  - .install
+  - .config
+  - .service
```
javierbertoli: `template/pkg/init.sls` shouldn't include `.config` nor `.service`, as no configuration or service management should happen under the `pkg` subdir, right?

template/pkg/service.sls
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
```
javierbertoli: `pkg` subdir should not have a service management state file. See that this service file should be moved to `template/service/init.sls`, as per [@noelmcloughlin ' s proposed structure](https://github.com/saltstack-formulas/template-formula/pull/48#pullrequestreview-207182085)

template/pkg/clean.sls
myii: Excellent catch, will fix that.

template/pkg/config.sls
myii: From the upstream-upstream documentation(!), i.e. the main Jinja docs.

http://jinja.pocoo.org/docs/2.10/templates/#math:

> `+`
> Adds two objects together. Usually the objects are numbers, but if both are strings or lists, you can concatenate them this way. **This, however, is not the preferred way to concatenate strings!** For string concatenation, have a look-see at the `~` operator. `{{ 1 + 1 }}` is 2.

http://jinja.pocoo.org/docs/2.10/templates/#other-operators:

> `~`
> Converts all operands into strings and concatenates them.
> 
> `{{ "Hello " ~ name ~ "!" }}` would return (assuming name is set to `'John'`) `Hello John!`.

* Emphasis above is mine.
* So in reality, the documentation you linked to needs to be updated.

javierbertoli: Yeah, I personally use `~` myself. Made the comment because read it in the docs and made sense to follow the doc. But you made more sense that the doc :smile: so let's go with `~`. Thanks!

template/pkg/init.sls
myii: My method of refactoring on community projects is to make the _smallest set of code changes possible_ while ensuring that there is no change in function.  So this was a direct translation of what `template/init.sls` was doing before this PR.

It is also based on how I've found existing formulas doing things, where installation-config-service are kept close together as a logical block.

However, your point is completely valid and represents the ideal.  A couple of questions before I add to this PR (and to leave time for others to comment):

1. We are still agreeing that running this formula should still perform all three stages, as it did before this PR, right?
1. In order to keep to maintain portability, how do you prefer that the other `sls` files are used in the `include` statements and the `require` requistes?
    1. My preference is _navigating from the root directory_ where `topdir` is resolved as [it is here](https://github.com/myii/template-formula/blob/52d03d895ac60cbcef64ca712881eee56d77c21a/template/map.jinja#L4-L10), for example:
        1. `{{ topdir }}.pkg.install`
        1. `{{ topdir }}.pkg.config`
        1. `{{ topdir }}.pkg.service`
    1. Another method is _relative navigation_, which starts by resolving the next directory up and going from there:
        1. `{{ parentdir }}.pkg.install`
        1. `{{ parentdir }}.pkg.config`
        1. `{{ parentdir }}.pkg.service`
 
That's not very clear.  Let me give examples for both.

Formula root directory-based navigation using `topdir`:

```sls
...
{%- set topdir = tpldir.split('/')[0] %}
{%- from topdir ~ "/map.jinja" import template with context %}
{%- from topdir ~ "/macros.jinja" import files_switch with context %}

include:
  - {{ topdir }}.pkg.install

template-config:
  ...
    - require:
      - sls: {{ topdir }}.pkg.install
```

Parent directory-based navigation using `parentdir` and `parentdot`:

```sls
...
{%- set parentdir = salt['file.dirname'](tpldir) %}
{%- set parentdot = parentdir|replace('/', '.') %}
{%- from parentdir ~ "/map.jinja" import template with context %}
{%- from parentdir ~ "/macros.jinja" import files_switch with context %}

include:
  - {{ parentdot }}.pkg.install

template-config:
  ...
    - require:
      - sls: {{ parentdot }}.pkg.install
```

template/pkg/service.sls
myii: As opened for discussion in the [comment above](https://github.com/saltstack-formulas/template-formula/pull/48#discussion_r259797184).

myii:
@javierbertoli Thanks for the excellent feedback.  It's given me some food for thought about wider implementation issues including some TOFSv2 stuff.  I now really believe that alongside `tplfile`, `tpldir` and `tpldot`, upstream should also provide `topdir` (or `rootdir`) as well by default.  That gives a fixed base for all path and dot-based navigation throughout entire formulas.

@aboe76 @noelmcloughlin Please see @javierbertoli's excellent comments above and my responses.  There are some important decisions to be made and your feedback will be invaluable.

myii:
@javierbertoli Can I trouble you to also look at [this comment above](https://github.com/saltstack-formulas/template-formula/pull/48#issuecomment-466814906)?  I know we discussed this in our Slack room but I can't recall what the final conclusion was.

template/pkg/init.sls
javierbertoli: > We are still agreeing that running this formula should still perform all three stages, as it did before this PR, right?

Yes, I think there's no discussion about this point: "the template formula should provide an example of:

* installing/removing a package (app), 
* configuring/deleting the config of such app
* starting/stopping (deleting) the service of such app.

My comment in this PR regarding the structure  is that you included the `config` and `service` states **inside** the `pkg` subdir.

```
            {{ topdir }}.pkg.install
            {{ topdir }}.pkg.config
            {{ topdir }}.pkg.service
```

But from @noelmcloughlin proposal (wich makes sense to me), the structure should be (from the formula's *topdir*)

```
{{ topdir }}.init.sls       # includes {pkg,config,service}.init
{{ topdir }}.clean.sls      # includes {service,config,pkg}.clean

{{ topdir }}.pkg            # `pkg/init.sls` which invokes `pkg.install`
{{ topdir }}.pkg.install
{{ topdir }}.pkg.clean

{{ topdir }}.config         # `config/init.sls` which configs the app
{{ topdir }}.config.clean

{{  topdir }}.service    # `service/init.sls which starts the service
{{topdir }}.service.clean
```

myii: @javierbertoli Yes, that's fine.  I pulled @noelmcloughlin into the conversation at the beginning because I knew he had interesting things to offer.

javierbertoli:
@myii I think testing for `clean` states *can* be done on the same test run. From the top of my head, I think it should involve creating a `clean` suite which requires first the `install` states. Perhaps like:

```yaml
suites:
  - name: default
  - name: clean
    provisioner:
      state_top:
        base:
          '*':
            - template       # Run all the "up" (install, config, service running) states
            - template.clean # Run all the "down" states
```

Then, a `test/integration/clean` subdir should be created, which verifies service is *not* running, config files do *not* exist and package is *not* installed.

I think that should do it. wdyt?

myii:
@javierbertoli That sounds right.  But how about this: how can we be 100% sure that than installation, configuration and service states even ran in the first place?!  Meaning, the tests that would be written here would match the pre-run state exactly.  What I really want to be able to do is:

1. Perform the installation-configuration-service states.
1. Run tests to ensure those have been completed successfully.
1. Perform the clean states.
1. Run tests to ensure those have been completed successfully.

That's the only way we can be 100% sure that the tests are passing.

javierbertoli:
> But how about this: how can we be 100% sure that than installation, configuration and service states even ran in the first place?

That's the point: the

```
suites:
  - name: default
```
already ensures the 'up' states run correctly and will do the *Right Thing (TM)*.

So, if the `default` suite fails, the formula needs to be fixed.

Now, regarding the `clean` suite, as the 1st state in the `clean` suite runs the 'up' state again, you're sure that the `clean` suit is preparing everything to test the `down` states. :smile:

The only cost is running the 'up' suite twice, but that's not much of a price, right?

myii:
@javierbertoli That's brilliant, you've put my mind at ease!  Of course, a little duplication isn't going to hurt.  So, can we look forward to your wonderful `clean` testing PR...?  This is the "price" of showing off your experience!  No pressure though, honest.

javierbertoli:
> So, can we look forward to your wonderful clean testing PR...? This is the "price" of showing off your experience! No pressure though, honest.

LOL!! Sweetening the deal, uh? If only I were that eas...OK, I'll do it. Although I foresee a busy week, I'll try to submit a PR these days :smile:

myii:
@javierbertoli You walked into that one!  But seriously, your offer is appreciated.  Take your time; we all look forward to another masterpiece of `inspec` testing!

template/pkg/init.sls
noelmcloughlin: I agree with @javierbertoli observations that configuration and service should be separate (from `pkg/ `) but we could capture those review comments for future PR.  Using '{{ topdir }}' instead of relative path has one advantage - in terms of consistency - where foreign states are included in the top states.   

For example the following highlights that `nginx`  and `{{ topdir }}` should co-exist in the top directory (i.e. base).

```
include:
  - nginx.ng
   - {{ topdir }}.pkg.install
```

This relationship *maybe* less obvious when parentdir is used.
```
include:
  - nginx.ng
   - {{ parentdir }}.pkg.install
```
 
However the scenario I am raising is not the common pattern so its minor consideration.

myii: @noelmcloughlin Thanks for the feedback.  I think we're reaching some good consensus decisions here.  I don't think it's a problem to add these changes to the PR and will do them shortly.  I'm glad that we're all preferring `topdir` -- it keeps things sane.  I'm seriously considering an upstream PR since this needs to be provided by default.

noelmcloughlin: One more thought (per the vision stuff, not this PR) is that salt can support both yaml and json configuration.  All formulas use `yaml/` today but anticipating `json/`  or other data serialisation technology language in the design may make sense - something like this?
```
   template-formula/
        template/
          .. etc ..
             yaml/
                  defaults.yaml
                  osmap.yaml
                  osfamilymap.yaml
            json:
                 defaults.json
                 osmap.json
                 osfamilymap.json
```

myii: @noelmcloughlin Yes, @aboe76 kicked of issue #24 for discussions about this.

myii:
CC: @aboe76 @javierbertoli @noelmcloughlin.

OK, that's everything done except for the `README`, which I'll push a following commit.  This has become quite a hefty change from what was submitted earlier so please look it over carefully.

myii:
I've just pushed the changes to the `README`, so the work here is done.

README.rst
```diff
@@ -67,3 +69,30 @@ via include list.
...
+
+This state will undo everything performed in the ``template`` meta-state in reverse order, i.e.
+stops the service,
+removes the configuration file and
+then uninstalls the package.
+
+``template.service.clean``
+--------------------------
+
+This state will stop the template service only.
```
javierbertoli: "This state will stop the service and disable it at boot time."

myii: No problem, will push now.

template/clean.sls
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
+
+include:
+  - .service.clean
+  - .config.clean
+  - .pkg.clean
```
noelmcloughlin: For consistency with `service` and `config` & `future` modules I suggest `pkg` be changed to `package`.

template/config/clean.sls
```diff
@@ -0,0 +1,17 @@
...
+{#- Get the `topdir` from `tpldir` #}
+{%- set topdir = tpldir.split('/')[0] %}
+{%- set sls_service_clean = topdir ~ '.service.clean' %}
+{%- from topdir ~ "/map.jinja" import template with context %}
+{%- from topdir ~ "/macros.jinja" import files_switch with context %}
+
+include:
+  - {{ sls_service_clean }}
+
+template-config-absent:
```
noelmcloughlin: For predictable API a useful naming convention is to align state names with formula structure - so `template-config-absent` becomes `template-config-clean-absent` or `template config clean absent` in natural language.  

The convention scales well and and is nice when looping:

```
{%- for x in list %}
template config clean {{ x }} absent:
{%- endfor %}
```

template/config/file.sls
```diff
@@ -21,3 +23,5 @@ template-config:
     - user: root
```
noelmcloughlin: Should `user` and `group` be configurable as general rule?   => future PR!!

template/config/init.sls
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
```
noelmcloughlin: How about line 1 being a breadcrumb?

```
1  # {{ tpldir }}/config/clean.sls
2 # # -*- coding: utf-8 -*-
```

template/pkg/init.sls
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
+
+include:
+  - .install
```
noelmcloughlin: Do we need `install.sls` because as general rule I prefer using `init.sls` for installation stuff to mirror fact that `clean.sls` removes stuff?  I would prefer `install.sls` be removed for the simple scenario.

template/pkg/install.sls
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
+
+{#- Get the `topdir` from `tpldir` #}
+{%- set topdir = tpldir.split('/')[0] %}
+{%- from topdir ~ "/map.jinja" import template with context %}
+
+template-pkg:
```
noelmcloughlin: More descriptive state name is `template pkg install pkg` per my previous comment.

template/service/init.sls
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
+
+include:
+  - .running
```
noelmcloughlin: Again`init.sls` could include the actual startup SLS so it mirrors `clean.sls` which manages the dead SLS.

template/service/running.sls
```diff
@@ -0,0 +1,17 @@
...
+
+{#- Get the `topdir` from `tpldir` #}
+{%- set topdir = tpldir.split('/')[0] %}
+{%- set sls_config_file = topdir ~ '.config.file' %}
+{%- from topdir ~ "/map.jinja" import template with context %}
+
+include:
+  - {{ sls_config_file }}
+
+template-service:
```
noelmcloughlin: More descriptive state name is `template service running enabled`.

template/pkg/init.sls
noelmcloughlin: Going back to your discussion with @javierbertoli regarding portability - portability can be maintained by redirecting legacy API (i.e.  old.sls) to an agreed/preferred API (i.e. new.sls) by simple include-

**old.sls**
```

include:
  - .new
```

This enforces depreciation of unwanted API for preferred API. The README must be updated accordingly to ONLY document preferred API - after a considered period of time (i.e. 12 months) the unwanted API can be ejected (i.e.  `cd template; rm old.sls; git add --all; git commit -m"remove depreciated SLS"; git push` )

template/clean.sls
myii: OK, that can be done.

template/config/clean.sls
myii: OK, I'll look at this as well.

template/config/file.sls
myii: We're trying to move away from pillar values as much as possible, so we need to replace these with one of the alternatives being looked into.  So leaving it for now, at least.

template/config/init.sls
myii: Even better:

```sls
# {{ tplfile }}
# -*- coding: utf-8 -*-
```

template/pkg/init.sls
myii: For consistency, let's leave all `init.sls` doing nothing more than directing traffic.  So it becomes easier to see the flow of the formula by opening up the `init.sls` files and seeing how they're linked.

myii: That's why I've also limited these files to use relative includes, so that they don't get abused for anything else.  Not even defining `topdir`.

template/pkg/install.sls
myii: OK.

myii: BTW, as this was just refactoring mainly, I didn't want to touch other lines.  But since we've all gone to so much effort here...

template/service/init.sls
myii: Here I see `running.sls` mirroring `clean.sls`.

template/service/running.sls
myii: OK.

myii:
@noelmcloughlin Thanks for the detailed dissection!  I'll push those amendments when I get a chance.  This has started from a relatively minor, obscure PR and has started to become a complete overhaul of the formula.  Talk about "feature creep"!  I'm wondering whether all of these changes shouldn't trigger a `major` release.  There's nothing left to change!!  I've touched everything!

template/pkg/init.sls
noelmcloughlin: okay, but following this logic `clean.sls` would needs to direct traffic too ;-)

template/config/init.sls
noelmcloughlin: nice

template/config/file.sls
noelmcloughlin: Cool. sounds reasonable.

myii:
Note, while preparing the [upstream PR](https://github.com/saltstack/salt/pull/51814) that's linked above, I decided that `tplroot` would be a more appropriate variable name as opposed to `topdir`.  That's why the last commit has made that change throughout.  _If_ that PR is merged, then none of these formulas would be affected.  After some time, the line `{%- set tplroot = tpldir.split('/')[0] %}` can simply be removed throughout formulas.

template/pkg/init.sls
myii: @noelmcloughlin I've linked both of your comments above to two of the open issues in this repo, #21 and #24.

template/config/clean.sls
myii: Done.

For a fully dependable, consistent structure, I saw 5 parts to the state ID naming:

* 3 parts from the filename:
    * `template/package/install.sls`
* 2 parts from the state:
    * `pkg.installed`
* In the structure:
    * `<formula>-<subdir1>-<subdir2>-<state1>-<state2>`
* Putting together to get:
    * `template-package-install-pkg-installed`

So across the whole formula:

| File                         | State           | State ID                                 |
|------------------------------|-----------------|------------------------------------------|
| template/package/install.sls | pkg.installed   | template-package-install-pkg-installed   |
| template/package/clean.sls   | pkg.removed     | template-package-clean-pkg-removed       |
| template/config/file.sls     | file.managed    | template-config-file-file-managed        |
| template/config/clean.sls    | file.absent     | template-config-clean-file-absent        |
| template/service/running.sls | service.running | template-service-running-service-running |
| template/service/clean.sls   | service.dead    | template-service-clean-service-dead      |

template/pkg/init.sls
noelmcloughlin: Thanks @myii

template/pkg/init.sls
noelmcloughlin: @aboe76 @javierbertoli  For subdirectories I prefer `directory/init.sls`  &  `directory/clean.sls` to include SLS states and do work so that init.sls and clean.sls mirror each other.

@myii prefers `init.sls` to exclude SLS and only have includes.  Do you guys have a preference?

template/config/init.sls
myii: OK, have spoken to @noelmcloughlin directly and we've decided that there's no need for this.

template/clean.sls
myii: Done.

README.rst
```diff
@@ -67,3 +69,30 @@ via include list.
 This state will start the template service and has a depency on ``template.config``
```
daks: s/depency/dependency

myii: @daks Good catch!  This has been a long PR...

template/pkg/init.sls
daks: I agree with `init.sls` only have includes

README.rst
myii: @daks Wasn't my fault, that was based on existing content!

myii:
CC: @javierbertoli @noelmcloughlin @aboe76 @daks.

All the remaining changes have been made so this is ready for final review.  One last decision needs to be made about the version bump.  @javierbertoli [pointed out](https://freenode.logbot.info/saltstack-formulas/20190225#c2025771) that perhaps this isn't a `major` release.  My opinion differs and I've included the `BREAKING CHANGE` block on both of the last commits (prior to the typo fix commit):

> BREAKING CHANGE: Changing the `pkg` directory to `package` will break
implementations that are depending on `pkg` for `include` or `sls`-based
requisite purposes.

> BREAKING CHANGE: Wholesale state ID changes will break implementations
that are relying on the previous state IDs for requisite purposes.

Other than what I've written above, I want to emphasise that when evaluating breaking changes, it is more than _just the formula itself_ that needs to be considered.  With all of the discussions that have been taking place recently, the plan is to find a suitable way to use these formulas in conjunction with one another.  So we _cannot assume that each formula is fully independent_.  So any changes that could cause breakages in the wild are breaking changes for the formula itself.

README.rst
myii: Fixed.

template/pkg/init.sls
aboe76: is this what formulas should look like? I'm hesitant about these changes,
@noelmcloughlin I liked the idea of moving the yaml stuff in a directory to keep the formula files clean, 
But having to many subfolders doing one thing makes it harder to implement this template design on the plus ``300``` formulas we already have...

Looking at how other config management tools have fixed this, maybe it is time to incoroporate those best practices:

```ansible```

Has a dedicated folder for defaults and for variables per os/os_family
has a dedicated template folder
has a dedicated file folder 
has a task folder where the main code goes that actually does stuff.

```puppet```

has a dedicated folder for main code (manifests)
has a dedicated folder for files and or templates
has lib folder structure for ruby plugins that can be used directly in the main code
can have hiera data yaml structures for variables.

```chef```
has a dedicated folder for everything attributes,definitions,files,libraries, providers, resources,templates and recipes

```saltstack```
has no dedicated folder for files or templates (arbitrary name)
has no dedicated folder for variables/defaults
has a dedicated folder for the main code where everything lives. (state name)

why because everything is data in saltstack, ans sls file can provide states, but also variables,
a jinja file can be a template but also a macro code block, and yaml files provide only data.
but salt can have json, mako,etc, almost any file type which hold data or structure...

I think it's near impossible to come up with a structured way, only @noelmcloughlin initial idea of having different solution templates, which other people can choose,

having a working model for digesting pillar or external data to a formula is one thing, but structuring the rest how states are build/modeled that highly depends on what the formula manages and to what extend.

**Comment by myii** [comment]
OK, there's more than one last decision to be made... see [this comment above](https://github.com/saltstack-formulas/template-formula/pull/48#discussion_r260441716).

File: template/pkg/init.sls (hunk 4 above)
Comment (daks) [reply template/pkg/init.sls]: @aboe76 Are you talking about the directory layout in this comment https://github.com/saltstack-formulas/template-formula/pull/48#pullrequestreview-207182085?

File: template/pkg/init.sls (hunk 11 above)
Comment (noelmcloughlin) [reply template/pkg/init.sls]: @daks  For consistency should `clean.sls` only have includes too?

File: template/pkg/init.sls (hunk 4 above)
Comment (daks) [reply template/pkg/init.sls]: I think you're talking about those changes. 
I personally feel that we are going from no best practices to a really complex setup with a lot of default directories. Some are fine, others are somewhat useless for most/simplest formulas.
What am I talking about? 

- the multiple yaml mapfiles (osfamily, osmap...) when in most cases you don't need them
- the yaml and json directories
- even the install/, config/, service/ directories

What I mean is: for most formulas is too much, and maybe a simpler layout could be sufficient for most of them. Though we should have a way to complexify this setup when needed.
I think that this complex layout can 'frighten' a new user.

Comment (aboe76) [reply template/pkg/init.sls]: @daks yes I'm talking about https://github.com/saltstack-formulas/template-formula/pull/48#pullrequestreview-207182085 

and the multiple yaml files are only to make the formulas portable across distros without too much hassle

Comment (daks) [reply template/pkg/init.sls]: I know. Maybe those (map.jinja and all os*.yaml) are the ones which should be in a sub-directory to lightened the layout

Comment (daks) [reply template/pkg/init.sls]: to ease separation of states and (OS) data

Comment (aboe76) [reply template/pkg/init.sls]: @daks thats true, and this is one of the things that needs to be discussed openly and maybe there are more solutions to the same problem.

Comment (myii) [reply template/pkg/init.sls]: @daks @aboe76 If this needs to be discussed in more detail, it will be far easier to do that in our Slack/IRC/Matrix room than via. GitHub comments...

Comment (javierbertoli) [reply template/pkg/init.sls]: @noelmcloughlin raised a need ('formulas should include 'tear down' capabilities) and with that suggested a solution (the directories structure and states).

While I agree the solution is complex, I OK'ed it because it was *better than what we have now* :smile: 

Another solution for that need is to keep out simple states:
```
{{ topdir }}.init.sls   # includes {pkg,config,service}
{{ topdir }}.package    #  manages pkg install/remove
{{ topdir }}.config     # `manages config/remove of config files
{{ topdir }}.service    # starts/stop/remove the service
```

and use a conditional to manage the desired state, like say, in pseudo for a pillar

```
template:
  pkg:
    ensure: absent  # default: installed
```

the state can be

```
template-package:
  pkg.{{ pkg.ensure }}
```

I used to do that a lot in puppet, and kept things simple. On the directory for the yaml/json files, I'd go with a *single* directory wich can have any of them, with a name like 'local_values' or something meaningful (sad that 'config' is taken :smile:)? and that's it?

My 2 GH cents ;)

Comment (noelmcloughlin) [reply template/pkg/init.sls]: @aboe76  If there are concerns we should stop and discuss! If nothing else, to capture your concerns because this PR can be superseded by another PR to address your concerns.   Beyond this PR there are many considerations that concern me.  

Here is my take - assume a formula consists of two parts-

**(1) state model**
The standard use case is a single **system** (i.e. `nginx`) with three components-
- system packages/libs,  
- system config
- system daemon

Three subdirectories models the system components and API calls-
```
template-formula/
    template/
           package/     (states:  template.package, template.package.clean)
           config/      (states: template.config, template.config.clean)
           service/     (states: template.service, template.service.clean)
```
**(2) data model**
Data artifacts are consumed by deployment automation model (Salt, but maybe Ansible, etc).  
We could try to figure out what directory structure works best.  
```
template-formula/
    template/
           software/package/     (states:  template.software.package, template.core.package.clean)
           software/config/         (states: template.software.config, template.core.config.clean)
           software/service/        (states: template.software.service, template.core.service.clean)
           model/yaml            
           model/jinja             (*.jinja, *.j2, templates, macros too)
           model/json
           model/mako
```

**Common formula patterns**

This list is not exhaustive.

* Docker runs the daemon
======
A docker container may provide the "service" but depend on `package` and `config` states for binding purposes (see https://docs.docker.com/storage/bind-mounts/). I have seen this pattern with nginx reverse proxy for example.

* Multiple services 
=====
The `iscsi-formula` is the best example I can think of.
```
iscsi-formula/
           isns/package/     
           isns/config/      
           isns/service/ 
           target/package/
           target/config/
           target/service/
           initiator/package/
           initiator/config/
           initiator/service
```
* Toolchains
======
A package state is needed, but the `service` state is not.  Examples are Java, vim, IntelliJ, etc.

* Cross Platform packaging
======
- Tarball 
- docker compose file
- git repo
- snap

I agree that this formula cannot handle everything but this needs discussion over longer period I guess.

myii:
CC: @javierbertoli @noelmcloughlin @aboe76 @daks.

Can we merge this and fix any structural changes or issues in future PRs?

daks:
@myii I'm OK with the merge, I think we agree on most of the code changed

myii:
@daks @noelmcloughlin Thank you both for your reviews (and all of the feedback along the way).

aboe76:
if everybody is on board with this change, then LGTM. and discuss it on slack/riot/irc

myii:
Thanks @aboe76.

@javierbertoli Would you be so kind as to hit the merge button?

PR #50, opened 2019-02-25
ci(kitchen+travis): disable `debian-8` due to `2019.2` installation bug

myii:
* https://github.com/saltstack/salt/issues/51808

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
 env:
   matrix:
     - INSTANCE: default-debian-9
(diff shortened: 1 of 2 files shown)


aboe76:
@myii debian-8 won't be fixed by saltstack team: https://github.com/saltstack/salt/issues/51808

we need to get an update debian 8.5 image and see if it works.

myii:
For Docker, using `debian:jessie-backports` has worked for us (the reference link directly above).

PR #52, opened 2019-02-26
ci(kitchen): check for repos updates before trying package installation

daks:
A little update for Kitchen provision_commands to update repository information before installing packages.
I had an old debian-9 container which couldn't install udev. This fixes it.

### DIFF
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -19,7 +19,7 @@ platforms:
     driver_config:
       run_command: /lib/systemd/systemd
       provision_command:
-        - apt install -y udev
+        - apt-get update && apt-get install -y udev
   # With the release of `2019.2`, `debian-8` is no longer working
   # Disabling until the following upstream bug has been resolved:
   #   * https://github.com/saltstack/salt/issues/51808
@@ -27,17 +27,17 @@ platforms:
   #   driver_config:
   #     run_command: /lib/systemd/systemd
   #     provision_command:
-  #       - apt install -y udev
+  #       - apt-get update && apt-get install -y udev
   - name: ubuntu-18.04
     driver_config:
       run_command: /lib/systemd/systemd
       provision_command:
-        - apt install -y udev
+        - apt-get update && apt-get install -y udev
   - name: ubuntu-16.04
     driver_config:
       run_command: /lib/systemd/systemd
       provision_command:
-        - apt install -y udev
+        - apt-get update && apt-get install -y udev
   - name: centos-7
     driver_config:
       image: centos:7
@@ -50,7 +50,7 @@ platforms:
     driver_config:
       run_command: /usr/lib/systemd/systemd
       provision_command:
-        - yum install -y udev
+        - yum -y update && yum -y install udev
   # As of February 2019, there have been problems getting `opensuse` to work:
   #   * `opensuse` is deprecated
   #   * `opensuse/leap` grabs `15.x`, which doesn't run the `inspec` tests
@@ -62,7 +62,7 @@ platforms:
       image: opensuse/salt-minion
       run_command: /usr/lib/systemd/systemd
       provision_command:
-        - zypper install -y udev
+        - zypper refresh && zypper install -y udev
         - systemctl enable sshd.service
         - cat /etc/os-release
 


myii:
@daks Thanks for this, I believe it is the right thing to do.  However, do you mind changing it from `apt-get` to `apt`?

https://debian-handbook.info/browse/stable/sect.apt-get.html:

> ## 6.2. aptitude, apt-get, and apt Commands
>
> APT is a vast project, whose original plans included a graphical interface. It is based on a library which contains the core application, and `apt-get` is the first front end — command-line based — which was developed within the project. `apt` is a second command-line based front end provided by APT which overcomes some design mistakes of `apt-get`.
>
> Numerous other graphical interfaces then appeared as external projects: `synaptic`, `aptitude` (which includes both a text mode interface and a graphical one — even if not complete yet), `wajig`, etc. The most recommended interface, `apt`, is the one that we will use in the examples given in this section. Note however that `apt-get` and `aptitude` have a very similar command line syntax. When there are major differences between `apt`, `apt-get` and `aptitude`, these differences will be detailed.

Key points from the above:

* `apt` is a second command-line based front end provided by APT which overcomes some design mistakes of `apt-get`.
* The most recommended interface, `apt`, is the one that we will use in the examples given in this section.

daks:
I switched to apt-get because of this:

```
sudo apt update|grep nginx

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

But I can change it back.

myii:
@daks Excellent point.  I will ping @aboe76 to decide how to proceed here but I think you're right.  After this PR, I was actually searching for a reference to preferring `apt` or `apt-get` in scripts and I didn't succeed.  So leave things as they are for now and this will probably be merged as-is.  Thanks for the contribution!

kitchen.yml
```diff
@@ -50,7 +50,7 @@ platforms:
     driver_config:
       run_command: /usr/lib/systemd/systemd
       provision_command:
-        - yum install -y udev
+        - yum check-update && yum install -y udev
```
myii: OK, this is causing a breakage:

```
STDERR: The command '/bin/sh -c yum check-update && yum install -y udev' returned a non-zero code: 100
---- End output of docker -H unix:///var/run/docker.sock build -f Dockerfile-kitchen20190226-5221-j1u7qd . ----
Ran docker -H unix:///var/run/docker.sock build -f Dockerfile-kitchen20190226-5221-j1u7qd . returned 100] on default-fedora
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
The command "bundle exec kitchen verify ${INSTANCE}" exited with 20.
```

This will need to be resolved before continuing.

javierbertoli:
I've been using `apt` for a while because I read some time ago, somewhere, that it is the recommended app and that it will replace `apt-get`. It merges the functionality of `apt-get +  apt-cache`.

All that saidl, in this case, I don't think it will be much of a difference to use one or another. 

My 2c.

aboe76:
I think testing before installing a package means that the container isn't maintained, and running behind in what the test are made for, checking if the formula does what is supposed to do on a clean up to date distribution...

that means catching errors early so updating is a good thing !! 
So I like it.

daks:
@javierbertoli I also use it daily, and like the combined functionality. The problem is only in non-interactive use.

@myii trying to debug the Fedora problem, it seems that `yum check-update` works in a shell but not in provision command. I don't really know how you are suppose to update your repos info on Fedora

myii:
@daks Found out what the problem is (but haven't checked for the solution yet).

http://man7.org/linux/man-pages/man8/yum.8.html:

> ##### check-update
> 
> Implemented so you could know if your machine had any updates
> that needed to be applied without running it interactively.
> **Returns exit value of 100 if there are packages available for
> an update**. Also returns a list of the packages to be updated
> in list format. Returns 0 if no packages are available for
> update. Returns 1 if an error occurred.  Running in verbose
> mode also shows obsoletes.

* Emphasis above is mine.
* Matches up with [our error earlier](https://github.com/saltstack-formulas/template-formula/pull/52#discussion_r260422455):
  - `... returned a non-zero code: 100`

So potential solutions:

1. Gobble the (`100`) exit code and continue with the next command.
1. Look at using `update` or `upgrade` from the manual page linked above.

daks:
@myii I found this repository https://github.com/fedora-cloud/Fedora-Dockerfiles and now use `yum -y update`, just waiting for checks to finish

myii:
@daks Thanks for your diligence!  Merged.

Issue #61, opened 2019-02-28
Add extension `.sls` or `.yml` to `pillar.example`


myii:
Suggested by @getSurreal on our Slack/IRC/Matrix room [from here](https://freenode.logbot.info/saltstack-formulas/20190227#c2028927):

> 22:47onslack
> <salt878> @myii With all the changes going on in template-formula, has anyone discussed renaming pillare.example to something like pillare_example.sls or .yml or something that will be recognized in an IDE?
> 22:48myii[m]
> No-one has discussed that.
> 22:49myii[m]
> Usually, you can set your IDE to recognise the file type.
> 22:53onslack
> <salt878> Of course that's true, just seems like an unnecessary workaround
> 22:58onslack
> <javier> never even considered it :P_if_ you decide that it's worth, then I'd vote for _pillar.example.yaml_?
> 23:17onslack
> <myii> @getSurreal Having thought about it, I think you're absolutely right. Why should anyone have to make any changes from their end? It should work out of the box.
> 23:17onslack
> <myii> @javierbertoli Since all of the pillars end up as `.sls` files, shouldn't that be `pillar.example.sls`?
> 23:18onslack
> <myii> @getSurreal Can you submit a PR or even an issue? You've got my backing and @javierbertoli's -- it's really just a matter of `.yaml` vs. `.yml` vs. `.sls`.
> 23:30onslack
> <javier> @myii didn't thought about the pillars' extension, just the content of the file :-)Any extension that make sense has my vote, so I'm ok with `pillar.example.sls` too

getSurreal:
@myii sorry I dropped off the slack feed and missed your request to create an issue.  

I did have another thought on this.  Isn't a '.' in the name bad form because salt interprets that as a directory.  I prefer .sls since that's the correct extension for pillar files.

pillarexample.sls
pillar_example.sls
pillar-example.sls  ( I'm leaning towards this one )

myii:
@getSurreal Since this file isn't processed by Salt, I'm easy either way.  This is a file rename in any case, so the final naming can be locked down when reviewing the PR.  I'm good with `pillar-example.sls` as well.  Are you happy to submit the PR?

getSurreal:
I understand the file isn't processed by Salt, but if it's bad form it sets a bad example.  
I have no problem submitting a PR.  I just thought I would wait a little for additional comments

myii:
[...]

PR #62, opened 2019-02-28
Issue 61 - Rename pillar.example to pillar-example.sls

getSurreal:
Proper extension for pillar files is .sls.  Not using pillar.example.sls because the extra period is bad form for Salt.


myii:
Closes #61.

---

As an introduction to this PR, please consider the following comments captured on the issue behind this PR:

* https://github.com/saltstack-formulas/template-formula/issues/61#issue-415389149
* https://github.com/saltstack-formulas/template-formula/issues/61#issuecomment-468125198
* https://github.com/saltstack-formulas/template-formula/issues/61#issuecomment-468130447

daks:
Good to me

noelmcloughlin:
Could it be named `pillar-example.yaml` instead? It does not & should not contain SLS code I guess

myii:
> Could it be named `pillar-example.yaml` instead? It does not & should not contain SLS code I guess

@noelmcloughlin The issue with this is that the convention with actual pillar files is to use `.sls`.

noelmcloughlin:
> > Could it be named `pillar-example.yaml` instead? It does not & should not contain SLS code I guess
> 
> @noelmcloughlin The issue with this is that the convention with actual pillar files is to use `.sls`.

Okay, SLS then.  I see `pillar.example` has qualifying comment: `# vim: ft=yaml`

myii:
Yes, it's not ideal that the file extension and content format diverge.  But even changing from `pillar.example` => `pillar-example.sls` may be a step too far, due to existing conventions.  Changing the extension of all pillar files to `.yml` may end up in mass rejection!

Out of interest, has anyone even tried accessing a pillar using `.yml`...?

myii:
As expected, it doesn't work for anything other than the `.sls` extension:

```salt
    Data failed to compile:
----------
    Pillar failed to render with the following messages:
----------
    Specified SLS 'template' in environment 'base' is not available on the salt master
```

noelmcloughlin:
i will consider raising bug. On salt i guess?  YAML is also data exchange
technology. So i should be able to migrate my yaml models between various
technology consumers ideally.

myii:
@noelmcloughlin Here's the code in [`salt/pillar/__init__.py`](https://github.com/saltstack/salt/blob/61baab97de511aab9273446062430f02623f5d2d/salt/pillar/__init__.py#L707-L725), which calls [`salt/fileclient.py`](https://github.com/saltstack/salt/blob/138c9321d43aa1aa3950df8509f3d47aa24fe244/salt/fileclient.py#L361-L374).  Somehow, I don't think this is going to be changed any time soon...

aboe76:
@myii , @getSurreal doesn't ```spm``` use the name pillar.example?

myii:
@aboe76 Yes, referenced literally in three places in [`salt/spm/pkgfiles/local.py`]:

1. [L82](https://github.com/saltstack/salt/blob/b44f0f1d0fd564690c6e45c6dd061c0e84d76def/salt/spm/pkgfiles/local.py#L82)
1. [L120](https://github.com/saltstack/salt/blob/b44f0f1d0fd564690c6e45c6dd061c0e84d76def/salt/spm/pkgfiles/local.py#L120)
1. [L148](https://github.com/saltstack/salt/blob/b44f0f1d0fd564690c6e45c6dd061c0e84d76def/salt/spm/pkgfiles/local.py#L148)

Even if there was a PR to fix this, it wouldn't be available for a long time.

noelmcloughlin:
Its officially a bug now:  https://github.com/saltstack/salt/issues/51910

waynew:
> Okay, SLS then. I see `pillar.example` has qualifying comment: `# vim: ft=yaml`

That's a hint for Vim to use yaml formatting, since `sls` isn't a known Vim filetype. If you have a problem with the discrepancy the correct place to file a bug is wherever Vim filetypes are implemented.

sls files are a *superset* of yaml. If you have a `.yaml` file that you just really want to trade between systems then you can simply do `ln -s yourfile.yaml somefile.sls` (at least I'm pretty sure Salt reads symlinks - if not you just have to swap those things around).

PR #77, opened 2019-03-11
refactor(tofs): avoid using “salt['config.get']” for formula writers

baby-gnu:
We can hide the call to “salt['config.get']” in the macro by only
asking for a namespace where to lookup the “source_files”.

* docs/TOFS_pattern.rst (Example): document the use of the new
  “namespace” parameter.

* template/macros.jinja: add a new “namespace” parameter.
  Lookup files override under
  “<tplroot>:tofs:sources_files:<namespace>” and fallback to
  “source_files” parameter.

* template/config/file.sls (template-config-file-file-managed): use
  the new “namespace” parameter

### DIFF
--- a/template/config/file.sls
+++ b/template/config/file.sls
@@ -13,12 +13,10 @@ include:
 template-config-file-file-managed:
   file.managed:
     - name: {{ template.config }}
-    - source: {{ files_switch(
-                    salt['config.get'](
-                        tplroot ~ ':tofs:source_files:template-config-file-file-managed',
-                        ['example.tmpl', 'example.tmpl.jinja']
-                    )
-              ) }}
+    - source: {{ files_switch(['example.tmpl', 'example.tmpl.jinja'],
+                              lookup='template-config-file-file-managed'
+                 )
+              }}
     - mode: 644
     - user: root
     - group: root
--- a/template/macros.jinja
+++ b/template/macros.jinja
@@ -1,4 +1,5 @@
 {%- macro files_switch(source_files,
+                       lookup=None,
                        default_files_switch=['id', 'os_family'],
                        indent_width=6) %}
   {#-
@@ -8,6 +9,8 @@
 
     Params:
       * source_files: ordered list of files to look for
+      * lookup: key under '<tplroot>:tofs:source_files' to override
+        list of source files
       * default_files_switch: if there's no pillar
         '<tplroot>:tofs:files_switch' this is the ordered list of grains to
         use as selector switch of the directories under
@@ -21,11 +24,8 @@
       Deploy configuration:
         file.managed:
           - name: /etc/yyy/zzz.conf
-          - source: {{ files_switch(
-                          salt['config.get'](
-                              tplroot ~ ':tofs:source_files:Deploy configuration',
-                              ['/etc/yyy/zzz.conf', '/etc/yyy/zzz.conf.jinja']
-                          )
+          - source: {{ files_switch(['/etc/yyy/zzz.conf', '/etc/yyy/zzz.conf.jinja'],
+                                    lookup='Deploy configuration'
                     ) }}
           - template: jinja
 
@@ -52,6 +52,11 @@
       tplroot ~ ':tofs:files_switch',
       default_files_switch
   ) %}
+  {#- Lookup files or fallback to source_files parameter #}
+  {%- set src_files = salt['config.get'](
+      tplroot ~ ':tofs:source_files:' ~ lookup,
+      source_files
+  ) %}
   {#- Only add to [''] when supporting older TOFS implementations #}
   {%- for path_prefix_ext in [''] %}
     {%- set path_prefix_inc_ext = path_prefix ~ path_prefix_ext %}
@@ -66,7 +71,7 @@
       {%- do fsl.append('') %}
     {%- endif %}
     {%- for fs in fsl %}
-      {%- for source_file in source_files %}
+      {%- for src_file in src_files %}
         {%- if fs %}
           {%- set fs_dir = salt['config.get'](fs, fs) %}
         {%- else %}
@@ -76,7 +81,7 @@
             path_prefix_inc_ext,
             files_dir,
             fs_dir,
-            source_file.lstrip('/')
+            src_file.lstrip('/')
         ]) %}
 {{ url | indent(indent_width, true) }}
       {%- endfor %}
--- a/docs/TOFS_pattern.rst
+++ b/docs/TOFS_pattern.rst
@@ -314,12 +314,10 @@ We can simplify the ``conf.sls`` with the new ``files_switch`` macro to use in t
      file.managed:
        - name: {{ ntp.config }}
        - template: jinja
-       - source: {{ files_switch(
-                     salt['config.get'](
-                         tplroot ~ ':tofs:source_files:Configure NTP',
-                         ['/etc/ntp.conf.jinja']
-                     )
-               ) }}
+       - source: {{ files_switch(['/etc/ntp.conf.jinja'],
+                                 lookup='Configure NTP'
+                    )
+                 }}
        - watch_in:
          - service: Enable and start NTP service
        - require:


baby-gnu:
Hello.

I think I can rewrite the commit message since it actually breaks things. Maybe there is a better way of doing it, it's open to suggestion.

Regards

myii:
@baby-gnu Just had a brief glance -- this looks like a very good idea.  What break things -- just the commit message or something else?

template/macros.jinja
```diff
@@ -52,6 +52,11 @@
       tplroot ~ ':tofs:files_switch',
       default_files_switch
   ) %}
+  {# Lookup files or fallback to source_files parameter #}
```
myii: Please use `{#-` instead of `{#`, so that it doesn't end up rendering a blank line:

```sls
    - source:

      - salt://template/files/ABC/example.tmpl
      - salt://template/files/Debian/example.tmpl
      - salt://template/files/default/example.tmpl
```

template/macros.jinja
```diff
@@ -66,7 +71,7 @@
       {%- do fsl.append('') %}
     {%- endif %}
     {%- for fs in fsl %}
-      {%- for source_file in source_files %}
+      {%- for source_file in files %}
```
myii: For consistency: `source_file` => `file`.  What do you think?

myii: If doing that, then change below: `source_file.lstrip('/')` => `file.lstrip('/')`.

myii: Update: Forget these comments, I'd prefer to solve [the comment above](https://github.com/saltstack-formulas/template-formula/pull/77/files#r264497818) instead.

myii:
@baby-gnu I've had a chance to look at this and test it out and I definitely agree that this is the right way to go forwards -- thanks for the contribution!  However, I'd just like a couple of things cleaned up, as I've mentioned in my comments above.  I also want to ask @aboe76 to have a look, so that we can be confident that we're heading in the right direction.

I've tested before and after with the following in my pillar (includes some extra tests for `files_switch`:

```yaml
  tofs:
    files_switch:
      - any/path/can/be/used/here
      - id
      - osfinger
      - os
      - os_family
    path_prefix: template_alt
    dirs:
      files: files_alt
      default: default_alt
    source_files:
      template-config-file-file-managed:
        - 'example_alt_source.tmpl'
        - 'example_alt.tmpl.jinja'
      formula.autofs.config.master:
        - 'auto_alt.master'
      formula.autofs.config.home:
        - 'auto_alt.homeLdap'
      formula.autofs.config.away:
        - 'auto_alt.awayLdap'
```

I've got the same results before and after:

```sls
    - source:
      - salt://template_alt/files_alt/any/path/can/be/used/here/example_alt_source.tmpl
      - salt://template_alt/files_alt/any/path/can/be/used/here/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/ABC/example_alt_source.tmpl
      - salt://template_alt/files_alt/ABC/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/Ubuntu-16.04/example_alt_source.tmpl
      - salt://template_alt/files_alt/Ubuntu-16.04/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/Ubuntu/example_alt_source.tmpl
      - salt://template_alt/files_alt/Ubuntu/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/Debian/example_alt_source.tmpl
      - salt://template_alt/files_alt/Debian/example_alt.tmpl.jinja
      - salt://template_alt/files_alt/default_alt/example_alt_source.tmpl
      - salt://template_alt/files_alt/default_alt/example_alt.tmpl.jinja

    - source:
      - salt://template_alt/files_alt/any/path/can/be/used/here/auto_alt.master
      - salt://template_alt/files_alt/ABC/auto_alt.master
      - salt://template_alt/files_alt/Ubuntu-16.04/auto_alt.master
      - salt://template_alt/files_alt/Ubuntu/auto_alt.master
      - salt://template_alt/files_alt/Debian/auto_alt.master
      - salt://template_alt/files_alt/default_alt/auto_alt.master

    - source:
      - salt://template_alt/files_alt/any/path/can/be/used/here/auto_alt.homeLdap
      - salt://template_alt/files_alt/ABC/auto_alt.homeLdap
      - salt://template_alt/files_alt/Ubuntu-16.04/auto_alt.homeLdap
      - salt://template_alt/files_alt/Ubuntu/auto_alt.homeLdap
      - salt://template_alt/files_alt/Debian/auto_alt.homeLdap
      - salt://template_alt/files_alt/default_alt/auto_alt.homeLdap

    - source:
      - salt://template_alt/files_alt/any/path/can/be/used/here/auto_alt.awayLdap
      - salt://template_alt/files_alt/ABC/auto_alt.awayLdap
      - salt://template_alt/files_alt/Ubuntu-16.04/auto_alt.awayLdap
      - salt://template_alt/files_alt/Ubuntu/auto_alt.awayLdap
      - salt://template_alt/files_alt/Debian/auto_alt.awayLdap
      - salt://template_alt/files_alt/default_alt/auto_alt.awayLdap
```

So I'm happy with this implementation.

template/macros.jinja
```diff
@@ -52,6 +52,11 @@
       tplroot ~ ':tofs:files_switch',
       default_files_switch
   ) %}
+  {# Lookup files or fallback to source_files parameter #}
+  {%- set files = salt['config.get'](
```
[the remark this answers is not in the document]
myii: Suggestion: `files` => `src_files`.

template/macros.jinja
myii: Suggestion: `file` => `src_file`, in both places.

template/config/file.sls
```diff
@@ -13,12 +13,10 @@ include:
...
   file.managed:
     - name: {{ template.config }}
-    - source: {{ files_switch(
-                    salt['config.get'](
-                        tplroot ~ ':tofs:source_files:template-config-file-file-managed',
-                        ['example.tmpl', 'example.tmpl.jinja']
-                    )
-              ) }}
+    - source: {{ files_switch(namespace='template-config-file-file-managed',
+                              source_files=['example.tmpl', 'example.tmpl.jinja']
```
myii: Using `namespace=` and `source_files=` seems to make sense at this stage.  However, I think it just makes this longer than in needs to be.  Not pushing this issue, just opening it for discussion.

baby-gnu:
> @baby-gnu Just had a brief glance -- this looks like a very good idea. What break things -- just the commit message or something else?

The `files_switch` API is broken, I introduced a new **first parameter**.

To make it compatible with users calling directly `salt['config.get']` we should invert the two parameters like:

```
{%- macro files_switch(source_files,
                       namespace,
                       default_files_switch=['id', 'os_family'],
                       indent_width=6) %}
```

This way it can be used as 

```
    - source: {{ files_switch(
                    salt['config.get'](
                        tplroot ~ ':tofs:source_files:template-config-file-file-managed',
                        ['example.tmpl', 'example.tmpl.jinja']
                    )
              ) }}
```

or
 
```
    - source: {{ files_switch(['example.tmpl', 'example.tmpl.jinja'],
                              namespace='template-config-file-file-managed',
                 )
              }}
```

It could even be missused (doing the lookup two times) with:

```
    - source: {{ files_switch(
                    salt['config.get'](
                        tplroot ~ ':tofs:source_files:template-config-file-file-managed',
                        ['example.tmpl', 'example.tmpl.jinja']
                    ),
                    namespace='template-config-file-file-managed',

              ) }}
```

So, feature or refactor? ;-)

template/config/file.sls
baby-gnu: Maybe `lookup=` is much more meanful than `namespace=`.

baby-gnu:
I rebased my work, integrating the different comments made.

The `default-debian-8-backports` test fail seems not related to my work.

I changed the new variable name, position and optionality.

People using the old format are not forced to change, they can still call `salt['config.get']` by themself:

```
    - source: {{ files_switch(
                    salt['config.get'](
                        tplroot ~ ':tofs:source_files:template-config-file-file-managed',
                        ['example.tmpl', 'example.tmpl.jinja']
                    )
                 )
              }}
```

or they can change and use the new optional `lookup` argument:

```
    - source: {{ files_switch(['example.tmpl', 'example.tmpl.jinja'],
                              lookup='template-config-file-file-managed'
                 )
              }}
```

There is no error if the formula writer mistakely added the `lookup` argument without changing the list of files:

```
    - source: {{ files_switch(
                    salt['config.get'](
                        tplroot ~ ':tofs:source_files:template-config-file-file-managed',
                        ['example.tmpl', 'example.tmpl.jinja']
                    ),
					lookup='template-config-file-file-managed'
                 )
              }}
```

Regards.

myii:
Restarted the failing tests and all OK from that end.

template/macros.jinja
myii: Thanks.

template/macros.jinja
myii: Thanks.

template/macros.jinja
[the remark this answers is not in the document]
myii: Thanks.

template/macros.jinja
```diff
@@ -7,6 +8,8 @@
     Files Switch (TOFS) pattern.
 
     Params:
+      * lookup: key under '<tplroot>:tofs:source_files' to override
+        list of source files
```
myii: _If_ we stick with this idea of using `lookup` as the second parameter, then this comment should be moved to under the `source_files` comment below.  The same order for both the parameters and the comments.

template/macros.jinja
```diff
@@ -52,6 +52,15 @@
       tplroot ~ ':tofs:files_switch',
       default_files_switch
   ) %}
+  {#- Lookup files or fallback to source_files parameter #}
+  {%- if not lookup %}
+    src_files = source_files
```
myii: Need this `set` in a Jinja block.

template/macros.jinja
```diff
@@ -52,6 +52,15 @@
...
+  {%- else %}
+    {%- set src_files = salt['config.get'](
+        tplroot ~ ':tofs:source_files:' ~ lookup,
+        source_files
+    ) %}
+  {%- endif %}
```
myii: This whole `if-else-endif` block is unnecessary, since the `config.get` will result in the default value of `source_files` being used.  If there's any discomfort with `lookup=None` as the top, this can be set to any other default value.  But it appears to work fine as it is.

template/macros.jinja
myii: But based on [the comment below](https://github.com/saltstack-formulas/template-formula/pull/77/files#r265373681), it will be removed anyway.

myii:
@baby-gnu I think this updated method is definitely an improvement -- thanks for the extra effort you've put in.  Please consider my new comments and let's wait for @aboe76 to review before merging.

template/macros.jinja
baby-gnu: @myii thanks, I made some tests and you are right, the `salt['config.get']` is done with `template:tofs:source_files:None` which has no result so the second argument is used.

template/macros.jinja
myii: Thanks.

template/macros.jinja
myii: Thanks.

template/macros.jinja
myii: Thanks.

myii:
@baby-gnu I'm really happy with this now -- an excellent refactor.  Thanks for your work and your patience.  Would you be willing to submit a follow-up PR to update [`docs/TOFS_pattern.rst`](https://github.com/saltstack-formulas/template-formula/blob/193bba805fc14d3f0f61c71b6ce4bbf45a60312c/docs/TOFS_pattern.rst)?

@aboe76 I'm happy with this to be merged, if you are OK with the changes as well.

baby-gnu:
> Would you be willing to submit a follow-up PR to update [`docs/TOFS_pattern.rst`](https://github.com/saltstack-formulas/template-formula/blob/193bba805fc14d3f0f61c71b6ce4bbf45a60312c/docs/TOFS_pattern.rst)?

I should have missed something @myii because I updated the example in `docs/TOFS_pattern.rst`.

myii:
@baby-gnu You're right, it has been updated!  I didn't look at the whole file, so I thought there may be more changes needed.

aboe76:
@baby-gnu and @myii, you are both right, this looks like an improvement and I like the `lookup` part, this is a better pattern to use. LGTM

myii:
@babilen5 As discussed on IRC/Matrix, list of formulas to be updated:

Formula | Assignees | Status
---|---|---
[systemd-formula](https://github.com/saltstack-formulas/systemd-formula/pull/27)||Small update
[lighttpd-formula](https://github.com/saltstack-formulas/lighttpd-formula/pull/4)||Small update
[stack-formula](https://github.com/claudekenni/stack-formula/pull/2)|@myii|Small update
[zabbix-formula](https://github.com/saltstack-formulas/zabbix-formula)||From v1
[varnish-formula](https://github.com/saltstack-formulas/varnish-formula)||From v1

alxwr:
So `macros.jinja` now becomes a library which can be just dropped inside any formula? :-) I'm all for that. Maybe we want a `libtofs.jinja`? I really like the idea of being able to update standard functionality by copy-pasting the macro files I need into specific formulas. Copy-paste keeps the workload low while avoiding inter-formula dependencies.

baby-gnu:
Hello.

> So macros.jinja now becomes a library which can be just dropped inside any formula? :-) I'm all for that. Maybe we want a libtofs.jinja?

I like the idea, but for [systemd-formula](https://github.com/saltstack-formulas/systemd-formula) it looks like there are some specific things to lookup files under the sub-directories for each component.

I did not took enough time to look at it but I think a `libtofs.jinja` is a great idea since it meaningful.

Regards.

myii:
@alxwr I really like the idea of a `libtofs.jinja`, keeping macros in separate files for ease of reuse.  @baby-gnu Your concerns for the [systemd-formula](https://github.com/saltstack-formulas/systemd-formula) are well-founded -- I've got some progress to share, so I'm going to merge this PR in the meantime.  I'll put something up and ping you all soon.  Thanks for all of your work, reviews and ideas.

PR #78, opened 2019-03-13
docs(tofs): incorrect path for “source_files” lookup key

baby-gnu:
* docs/TOFS_pattern.rst: rename “nfs” to “ntp”.


myii:
@baby-gnu Thanks for that, merged.

PR #79, opened 2019-03-21
refactor(tofs): move “files_switch” macro to “libtofs.jinja”

baby-gnu:
BREAKING CHANGE: every formula writer will need to change the import
to use this new version.

* template/libtofs.jinja: provides the “files_switch” macro.

* docs/TOFS_pattern.rst: update documentation to use the new path.

* template/config/clean.sls: change import from “macros.jinja” to “libtofs.jinja”.

* template/config/file.sls: ditoo.


myii:
@baby-gnu Thanks for this -- merged.

PR #83, opened 2019-03-24
refactor(tofs): update with latest changes inc. 100% portable `libtofs.jinja`

myii:
Effectively identical to https://github.com/saltstack-formulas/systemd-formula/pull/27, except that (A) has been provided by merged PR #79.

PR #84, opened 2019-03-25
refactor(tofs): use `config` rather than `pillar` throughout

baby-gnu:
Import from `systemd-formula`.


baby-gnu:
Duplicate pull request with #83

PR #85, opened 2019-03-25
refactor(tofs): ensure (v2 > v1 > default) checking for `src_files`

baby-gnu:
Import from `systemd-formula`.


baby-gnu:
Duplicate with #83

PR #88, opened 2019-04-07
test(saltcheck) Add Basic Saltcheck Tests

claudekenni:
From the recent discussions about Saltcheck, here are some tests.
These should do the same thing as the Inspec Tests

see: https://saltstackcommunity.slack.com/archives/C7LG8SV54/p1554534766188900

One Issue is that Saltcheck does not seem to work with relative paths, importing data from map.jinja throws TemplateNotFound for the yaml files
```
salt.exceptions.SaltRenderError: Jinja error: ./osfamilymap.yaml
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/utils/templates.py", line 381, in render_jinja_tmpl
    output = template.render(**decoded_context)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 989, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "<template>", line 2, in top-level template code
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 1033, in make_module
    return TemplateModule(self, self.new_context(vars, shared, locals))
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 1090, in __init__
    self._body_stream = list(template.root_render_func(context))
  File "/var/cache/salt/minion/files/base/template/map.jinja", line 8, in top-level template code
    {%- import_yaml tplroot ~ "/osfamilymap.yaml" or {} as osfamilymap %}
  File "/usr/lib/python2.7/site-packages/salt/utils/jinja.py", line 140, in get_source
    raise TemplateNotFound(template)
TemplateNotFound: ./osfamilymap.yaml
```

### DIFF
--- a/template/saltcheck-tests/init.tst
+++ b/template/saltcheck-tests/init.tst
@@ -1,0 +1,38 @@
+{% for key,value in {
+  'mode': '0644',
+  'user': 'root',
+  'group': 'root',
+}.items() %}
+template-config-file-file-managed-test-{{ key }}:
+  module_and_function: file.get_{{ key }}
+  args:
+    - /etc/template-formula.conf
+  assertion: assertEqual
+  expected-return: '{{ value }}'
+{% endfor %}
+
+template-config-file-file-managed-test-contents:
+  module_and_function: file.search
+  args:
+    - /etc/template-formula.conf
+    - 'This is an example file from SaltStack template-formula.'
+  assertion: assertTrue
+
+template-package-install-pkg-installed-test:
+  module_and_function: pkg.version
+  args:
+    - bash
+  assertion: assertNotEqual
+  expected-return: ''
+
+template-service-running-service-status:
+  module_and_function: service.status
+  args:
+    - systemd-udevd
+  assertion: assertTrue
+
+template-service-running-service-enabled:
+  module_and_function: service.status
+  args:
+    - systemd-udevd
+  assertion: assertTrue


mchugh19:
Can you paste a snippet of the relative import syntax. This seems to work for me:
```
{% from "ntp/map.jinja" import ntp with context %}

check_installed:
  module_and_function: pkg.version
  args:
    - {{ ntp.client }}
  assertion: assertNotEmpty

{# check that our custom ntp server exists in the ntp.conf file #}
ensure_correct_file:
  module_and_function: file.contains
  args:
    - {{ ntp.ntp_conf }}
    - '0.ubuntu.pool.ntp.org'
  assertion: assertTrue

check_ntp_running:
  module_and_function: service.status
  args:
    - {{ ntp.service }}
  assertion: assertTrue

validate_user:
  module_and_function: user.info
  args:
    - root
  assertion: assertNotEmpty
  assertion_section: shell
```

mchugh19:
I should also mention, we might want to aim for using the neon version of saltcheck. It contains additional asserts and functionality.

template/saltcheck-tests/init.tst
```diff
@@ -0,0 +1,38 @@
...
+  args:
+    - /etc/template-formula.conf
+    - 'This is an example file from SaltStack template-formula.'
+  assertion: assertTrue
+
+template-package-install-pkg-installed-test:
+  module_and_function: pkg.version
+  args:
+    - bash
+  assertion: assertNotEqual
```
mchugh19: This could also be assertNotEmpty

claudekenni: > Can you paste a snippet of the relative import syntax. This seems to work for me:

With the newest saltcheck version from develop the imports worked.

daks:
Just a couple of remarks:

- saltcheck uses salt code itself to do its jobs. I'm not completely confortable with a tool which is 'judge and judged'
- saltcheck is only available in a yet-to-be-announced release. Which means is not available for 2018.3 or 2019.2 which are the supported releases
- one advantage is a reuse of jinja maps and data. It's can be quite interesting, even if, for better formulas tests readability, I'm not sure if it's an advantage or an inconvenient. Moreover it's also possible to inject this data in Inspec (we do it at work)
- if Inspec solves our needs and it's already implemented in part of existing formulas, why do we want to change it? for what seems a less mature tool? _(NIH syndrom?)_

mchugh19:
> * saltcheck uses salt code itself to do its jobs. I'm not completely confortable with a tool which is 'judge and judged'

I don't see an issue here and I think there are benefits to using the same tool. 
- These tests are testing the results of salt states. I wouldn't generally expect states to fail silently, but even so, saltcheck uses salt modules for data gathering. So it isn't necessarily the same tool which is configuring the state of a machine, and also validating it. Short of major bugs, I don't know of a way for a salt module to fail open and match an assert when it shouldn't. 
- By using saltcheck you have a close to state syntax. For some shops (including ones I've worked in) we didn't have the luxury of highly versatile engineers capable of juggling multiple tools and syntax to perform their jobs. While this is very much a people and culture problem, saltcheck helps reduce the cognitive burden to teams trying to write tests.
- By using saltcheck, you do not require another tool to be installed. While this PR is generally attempting to handle the use-case for formula testing as part of a development/CI workflow, using saltcheck also allows you to execute tests on your deployed hosts. We've used this as a troubleshooting step to verify that a user's account and ssh keys were correct and up to date as a step 1 automatic ticketing step. Due to other items in your list, it may not be perfect for the CI case, it wouldn't be bad to host saltcheck tests upstream as it would give users of the formulas a quick way to do their own validation ("Is my monitoring agent deployed correctly?" `salt <host> saltcheck.run_state_tests monitoring_agent`)
- Also, and again not necessarily applicable to PR, you can write saltcheck tests to assist in QA or environment validation. We used this to ensure that hadoop clusters were fully provisioned, with applications and configs correctly deployed, and inter-node communication working. In the same way that these formulas are built with an expectation for local config changes, we could allow for local saltcheck tests for local deployment customizations. 


> * saltcheck is only available in a yet-to-be-announced release. Which means is not available for 2018.3 or 2019.2 which are the supported releases

It was released in 2018.3, but as previously mentioned, the version with Neon is much improved. So basically, yep! Agreed!

> * one advantage is a reuse of jinja maps and data. It's can be quite interesting, even if, for better formulas tests readability, I'm not sure if it's an advantage or an inconvenient. Moreover it's also possible to inject this data in Inspec (we do it at work)

Saltcheck tests are run through the salt renderer allowing for pillar or map.jinja access, but just as easily you could host data independently in local jinja variables or otherwise hardcoding expectations. Since it can do both, it has all the advantages! As for inspec data sourced from jinja maps or pillars, can you share that work? It sounds cool!

> * if Inspec solves our needs and it's already implemented in part of existing formulas, why do we want to change it? for what seems a less mature tool? _(NIH syndrom?)_

Agreed! If inspec is rocking it, everyone understands it, and it works well, then you shouldn't change. But I'd say there is value in adding saltcheck tests as they can still be leveraged by users who only have salt, want to run tests in their environment, or have things to check which can't be done in inspec (the cron formula for example doesn't have inspec tests for me to see if inspec does do it, but the saltcheck tests are validating all the pillar config; not just the service.installed and service.running, but each minute, hour, day, etc attribute of each cron config)

Thanks for the thoughts @daks

n-rodriguez:
Hi!

I agree with @daks :

> * saltcheck uses salt code itself to do its jobs. I'm not completely confortable with a tool which is 'judge and judged'

Me neither.

> * saltcheck is only available in a yet-to-be-announced release. Which means is not available for 2018.3 or 2019.2 which are the supported releases

This sounds like a poor solution. It forces people to move forward with Salt releases which is not always possible just to run tests.

> * For some shops (including ones I've worked in) we didn't have the luxury of highly versatile engineers capable of juggling multiple tools and syntax to perform their jobs.

> using saltcheck also allows you to execute tests on your deployed hosts

You can also achieve this with Inspec.

mchugh19:
Thanks for the feedback @n-rodriguez! 

> > * saltcheck uses salt code itself to do its jobs. I'm not completely confortable with a tool which is 'judge and judged'
> 
> Me neither.

Why is that? As I tried to explain, these are actually different tools doing the configuration (salt states) and validation (salt modules). Is there a failure mode you are trying to detect? 
There might also benefit to sharing data between the two and thus having the capability to validate every setting. 
 
> > * saltcheck is only available in a yet-to-be-announced release. Which means is not available for 2018.3 or 2019.2 which are the supported releases
> 
> This sounds like a poor solution. It forces people to move forward with Salt releases which is not always possible just to run tests.

Agreed, this isn't great. Assuming the Neon release works for people, I wouldn't expect many changes in the future since it's really the salt modules that do all the work. Also wouldn't we be having a similar conversation about migrating tests to future inspec releases too?

> > * For some shops (including ones I've worked in) we didn't have the luxury of highly versatile engineers capable of juggling multiple tools and syntax to perform their jobs.
> 
> > using saltcheck also allows you to execute tests on your deployed hosts
> 
> You can also achieve this with Inspec.

But the difference is an environment provisioned by salt will have the bits to run saltcheck, (likely) the ability to run it over remote execution in bulk, and a system for processing the return. In our environments, for example, we don't have ruby. So just installing inspec is already a large ask. 


I'm not saying that saltcheck is superior to inspec. Just that for my use-cases it does what I need, and like the rest of salt, found it a great platform for solving problems. We've been able to take engineers who can build salt states, and got them creating saltcheck tests in no time. We've also used it as part of a post build validation of large hadoop clusters. Because of that, I thought it beneficial to share more with the community, but if it's not the appropriate tool for this job, that's no problem. But even if that's the case, I think you could consider allowing saltcheck tests upstream to allow greater testing/validation for the users that can use it as well as more examples of cool salt stuff can do.

daks:
> Thanks for the feedback @n-rodriguez!
> 
> > > * saltcheck uses salt code itself to do its jobs. I'm not completely confortable with a tool which is 'judge and judged'
> > 
> > 
> > Me neither.
> 
> Why is that? As I tried to explain, these are actually different tools doing the configuration (salt states) and validation (salt modules). Is there a failure mode you are trying to detect?
> There might also benefit to sharing data between the two and thus having the capability to validate every setting.

What I'm thinking about is the following:
- saltcheck is part of salt, you found the code at https://github.com/saltstack/salt/blob/develop/salt/modules/saltcheck.py
- it uses salt code to do its job of 'testing the logic of states and highstates', looking at https://github.com/saltstack/salt/blob/develop/salt/modules/saltcheck.py#L200 we see it imports salt libraries

If we imagine the situation where:
- a salt state is buggy, let's say `service.running`. The bug is that the state does not verify correctly the 'state' of the service, it see it running while it's not, so it does not run the exec module to start the service
- we use saltcheck to verify that our code runs like needed
- saltcheck uses the same code than salt to verify the state of the service: it suffers the same bug

I don't know if it's the case, if bugs in salt would affect saltcheck, but I can imagine it could be the case.

Also this situation (the bug) may be quite unlikely but I still prefer to be sure it won't affect me, that's why I said "I'm not completely confortable with a tool which is 'judge and judged'". I prefer to use a completely different and separate tool to verify the result of Salt execution.

---

Also, about
> one advantage is a reuse of jinja maps and data. It's can be quite interesting, even if, for better formulas tests readability, I'm not sure if it's an advantage or an inconvenient. Moreover it's also possible to inject this data in Inspec (we do it at work)

We do it this way:

- after salt execution _(kitchen converge)_, we extract pillars with a command like `inspec exec <platform> "-c 'sudo salt-call --local --id #{@minion_id} --config /tmp/kitchen/etc/salt --out json --log-level quiet pillar.items'"` and write the output in a temp directory
- in inspec, we add a `config.rb` library in the `libraries` directory, near the `controls` one. This Ruby code read the dump file and make data available
- then in all `controls` we use code like `apt_config = config(self).dig('pillar', 'apt', 'repositories') || {}` to retrieve specific data from the library

myii:
I've revisited this and have had encouraging results.  Because of the issues encountered, the process was quite involved, so I'll report back the findings in multiple comments rather than one monolithic one.

The reality is that we weren't in a position to work with this six months back.  Since then, our CI setup has matured sufficiently to be able to implement this and keep it maintained:

1. We now have pre-salted `develop` images that each formulas is testing against; `saltchecker` works reliably there.
1. We now have ~50 formulas that are using `semantic-release`, that are all managed by the `ssf-formula` -- so any changes are propagated across them all, customised per formula where the need arises.

myii:
> Can you paste a snippet of the relative import syntax.

@mchugh19 The issue here is that `saltcheck` doesn't bring in any of the `tpldata` values.

https://github.com/saltstack/salt/pull/51814

> | `tpldata` | without context    | with context             |
> |-----------|--------------------|--------------------------|
> | `tplfile` | template/map.jinja | template/pkg/install.sls |
> | `tpldir`  | template           | template/pkg             |
> | `tpldot`  | template           | template.pkg             |
> | `tplroot` | template           | template                 |

* See the files changed in the PR for the underlying code.

So using the following snippet in the test file, to log what's going on:

```sls
{%- do salt.log.error('tplroot\n' ~ tplroot) %}
{%- do salt.log.error('tplfile\n' ~ tplfile) %}
{%- do salt.log.error('tpldir\n' ~ tpldir) %}
{%- do salt.log.error('tpldot\n' ~ tpldot) %}
{%- do salt.log.error('slspath\n' ~ slspath) %}
```

Led to this problem, before a workaround was found.

https://travis-ci.org/myii/template-formula/jobs/597282982#L1396-L1421

```sls
       [ERROR   ] tplroot
       .
       [ERROR   ] tplfile
       
       [ERROR   ] tpldir
       .
       [ERROR   ] tpldot
       
       [ERROR   ] slspath
       
       [ERROR   ] Rendering exception occurred
       Traceback (most recent call last):
         File "/usr/lib/python3/dist-packages/salt/utils/templates.py", line 393, in render_jinja_tmpl
           output = template.render(**decoded_context)
         File "/usr/lib/python3/dist-packages/jinja2/asyncsupport.py", line 76, in render
           return original_render(self, *args, **kwargs)
         File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 1008, in render
           return self.environment.handle_exception(exc_info, True)
         File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 780, in handle_exception
           reraise(exc_type, exc_value, tb)
         File "/usr/lib/python3/dist-packages/jinja2/_compat.py", line 37, in reraise
           raise value.with_traceback(tb)
         File "<template>", line 7, in top-level template code
         File "/usr/lib/python3/dist-packages/salt/utils/jinja.py", line 172, in get_source
           raise TemplateNotFound(template)
       jinja2.exceptions.TemplateNotFound: ./map.jinja
```

The workaround involves using a local copy of `map.jinja`, which doesn't rely on any of the `tpldata` values.  So for the `template-formula`, a simply find/replace solution is:

```diff
--- template/map.jinja
+++ template/saltcheck-tests/map.jinja
@@ -1,21 +1,19 @@
 # -*- coding: utf-8 -*-
 # vim: ft=jinja
 
-{#- Get the `tplroot` from `tpldir` #}
-{%- set tplroot = tpldir.split('/')[0] %}
 {#- Start imports as #}
-{%- import_yaml tplroot ~ "/defaults.yaml" as default_settings %}
-{%- import_yaml tplroot ~ "/osarchmap.yaml" as osarchmap %}
-{%- import_yaml tplroot ~ "/osfamilymap.yaml" as osfamilymap %}
-{%- import_yaml tplroot ~ "/osmap.yaml" as osmap %}
-{%- import_yaml tplroot ~ "/osfingermap.yaml" as osfingermap %}
+{%- import_yaml "template" ~ "/defaults.yaml" as default_settings %}
+{%- import_yaml "template" ~ "/osarchmap.yaml" as osarchmap %}
+{%- import_yaml "template" ~ "/osfamilymap.yaml" as osfamilymap %} 
+{%- import_yaml "template" ~ "/osmap.yaml" as osmap %}
+{%- import_yaml "template" ~ "/osfingermap.yaml" as osfingermap %} 
 
 {#- Retrieve the config dict only once #}
-{%- set _config = salt['config.get'](tplroot, default={}) %}
+{%- set _config = salt['config.get']("template", default={}) %}
 
 {%- set defaults = salt['grains.filter_by'](
       default_settings,
-      default=tplroot,
+      default="template",
       merge=salt['grains.filter_by'](
         osarchmap,
         grain='osarch',
```

And then use the local `map.jinja` instead in `template/saltcheck-tests/init.tst`:

```sls
{%- from "template/saltcheck-tests/map.jinja" import template with context %}
```

myii:
@mchugh19 The other problem with `saltcheck` is that it doesn't send an appropriate exit code upon failure.

https://travis-ci.org/myii/template-formula/jobs/597306680#L1541-L1553

```sls
             TEST RESULTS:
          ----------
          Execution Time:
              3.559
          Failed:
              1
          Missing Tests:
              0
          Passed:
              6
          Skipped:
              0
The command "bin/kitchen exec "${INSTANCE}" -c "sudo salt-call --config-dir=/tmp/kitchen/etc/salt saltcheck.run_highstate_tests"" exited with 0.
```

This was easily remedied by using [Retcode Passthrough](https://github.com/saltstack/salt/blob/01b9405b61cd416d4c852c87bd484759f5ec9c96/doc/topics/return_codes/index.rst#retcode-passthrough).

https://github.com/myii/salt/commit/e7a48c0427c1780696f1be2f9d6192b226bc6fde

```diff
diff --git a/salt/modules/saltcheck.py b/salt/modules/saltcheck.py
index 27d1f4f0316a..22c2bad874f8 100644
--- a/salt/modules/saltcheck.py
+++ b/salt/modules/saltcheck.py
@@ -382,6 +382,7 @@ def _generate_out_list(results):
     out_list.append({'TEST RESULTS': {'Execution Time': round(total_time, 4),
                                       'Passed': passed, 'Failed': failed, 'Skipped': skipped,
                                       'Missing Tests': missing_tests}})
+    __context__['retcode'] = (1 if (failed + skipped + missing_tests) else 0)
     return out_list
```

* Obviously, there may be some dispute about whether all of `failed + skipped + missing_tests` should result in an error or not.

In the meantime, I'm injecting this in during the Travis run and getting the desired results:

* https://travis-ci.org/myii/template-formula/jobs/597387780#L1545-L1556

This is working for all of our `develop` instances:

* https://travis-ci.org/myii/template-formula/builds/597387779 [with a failing `saltcheck` test]
* https://travis-ci.org/myii/template-formula/builds/597390289 [with passing `saltcheck` tests]

myii:
@claudekenni I had to update the tests in this PR to work with the latest changes to the formula.

```diff
@@ -1,5 +1,5 @@
 {%- set tplroot = tpldir.split('/')[0] %}
-{%- from tplroot ~ "/map.jinja" import template with context %}
+{%- from "template/saltcheck-tests/map.jinja" import template with context %}
 
 {% for key,value in {
   'mode': '0644',
@@ -18,13 +18,13 @@
   module_and_function: file.search
   args:
     - {{ template.config }}
-    - 'This is an example file from SaltStack template-formula.'
+    - 'This is another example file from SaltStack template-formula.'
   assertion: assertTrue
 
 template-package-install-pkg-installed-test:
   module_and_function: pkg.version
   args:
-    - {{ template.package }}
+    - {{ template.pkg.name }}
   assertion: assertNotEmpty
 
 template-service-running-service-status:
```

Would you mind making those adjustments (including bringing in the local version of `map.jinja`)?  If not, I'll push them to this PR myself, as it progresses.

PR #92, opened 2019-04-23
ci(kitchen): use pre-salted images instead

myii:
@javierbertoli has prepared pre-salted images, which run much faster.  This `WIP` PR implements these for CI testing instead.

---

- [x] Confirm this implementation is OK -- updated from @javierbertoli's gist [linked below](https://github.com/saltstack-formulas/template-formula/pull/92#issuecomment-485759607).
- [x] Remove all `udev` installations from `provision_command` once the images have that installed.
- [x] Decide what to do about `centos-6`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,20 +18,16 @@ before_install:
 # the `platforms` defined in `kitchen.yml`
 env:
   matrix:
-    - INSTANCE: v2019-2-py3-debian-9
-    - INSTANCE: v2019-2-py3-ubuntu-1804
-    - INSTANCE: v2019-2-py2-centos-7
-    - INSTANCE: v2019-2-py2-fedora-29
-
-    - INSTANCE: v2018-3-py2-debian-8
-    - INSTANCE: v2018-3-py2-ubuntu-1604
-    # - INSTANCE: v2018-3-py2-bootstrap-centos-6
-    - INSTANCE: v2018-3-py2-forced-version-fedora-28
-    - INSTANCE: v2018-3-py2-opensuse-423
-
-    - INSTANCE: v2017-7-py2-debian-8
-    - INSTANCE: v2017-7-py2-ubuntu-1604
-    # - INSTANCE: v2017-7-py2-bootstrap-centos-6
+    - INSTANCE: default-debian-9-2019-2-py3
+    - INSTANCE: default-ubuntu-1804-2019-2-py3
+    - INSTANCE: default-centos-7-2019-2-py2
+    - INSTANCE: default-fedora-29-2019-2-py2
+    - INSTANCE: default-opensuse-423-2018-3-py2
+    - INSTANCE: default-debian-8-2018-3-py2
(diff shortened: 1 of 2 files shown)


javierbertoli:
@myii , fixed the scripts, added the missing deps and uploaded new images.

I left [here](https://gist.github.com/javierbertoli/462671a3bcc9d5df76a0dd88e6042df4) cleaned up `kitchen.yml` and `.travis.yml` files, merging what you did here and what I was prepraring. I tested them locally and they 'work for me' (TM) :yum:

Regarding `centos-6`, as I was building locally, I cannot test it due to the [vsyscall issue](https://github.com/saltstack-formulas/template-formula/blob/master/kitchen.yml#L46-L47).

But when the packer scripts are ready to build working images, I don't see why not to build the 'complete' matrix (os * os_version * salt_version * python_version) that 'makes sense'? We can discuss this in the irc/slack channels, wdyt?

myii:
> @myii , fixed the scripts, added the missing deps and uploaded new images.

@javierbertoli Great work!  So it should be pulling the new images automatically, right?  Nothing else needs configuring here, does it?

> I left [here](https://gist.github.com/javierbertoli/462671a3bcc9d5df76a0dd88e6042df4) cleaned up `kitchen.yml` and `.travis.yml` files, merging what you did here and what I was prepraring. I tested them locally and they 'work for me' (TM) yum

The link you've provided ends in a `404` -- do you have an updated link?

> Regarding `centos-6`, as I was building locally, I cannot test it due to the [vsyscall issue](https://github.com/saltstack-formulas/template-formula/blob/master/kitchen.yml#L46-L47).
> 
> But when the packer scripts are ready to build working images, I don't see why not to build the 'complete' matrix (os * os_version * salt_version * python_version) that 'makes sense'? We can discuss this in the irc/slack channels, wdyt?

Oh dear, Travis isn't going to like you!  But that sounds very interesting indeed.  Look forward to these discussions.

javierbertoli:
Sorry, the correct link is to the gist is https://gist.github.com/javierbertoli/2475fb8ba1ffeffcde62226357694e7a

myii:
Much cleaner implementation in the gist so I've used that as the commit instead, authored by @javierbertoli.  Just going to confirm that the final tests are successful and then I'll merge this.

daks:
LGTM :)

daks:
Only remark _(the same as before I think)_: how to we industrialize creation of images? Are those run on travis? Is the code to generate them available in the saltstack github org? ...

myii:
@daks I may be wrong but I believe @javierbertoli intends to share the scripts and even maybe have these within the organisation itself.

daks:
I found the repo digging in slack backlog https://github.com/netmanagers/salt-image-builder/blob/master/salt-testing-docker.json :)

We a

myii:
Good work @javierbertoli and thanks for the review and merge @daks!

PR #60, opened 2019-02-27
docs(yaml): os*.yaml map files needs at least an empty dict

daks:
fixes #59 with a little inline comment in specified files.

The alternative would be to document how to remove unused import from map.jinja, but the doc about 'template-formula use' is still to be done, so I prefer to add this little comment meanwhile.

Issue #59, opened 2019-02-27
Inline comments for os*.yaml files is wrong, can't be empty


daks:
It seems that if any of the os*.yaml files is empty, salt crashes with
```
       {%- import_yaml topdir ~ "/osmap.yaml" as osmap %}
       {%- import_yaml topdir ~ "/osfingermap.yaml" as osfingermap %}
 
       {% set defaults = salt['grains.filter_by'](default_settings,    <======================
           default='template',
           merge=salt['grains.filter_by'](osfamilymap, grain='os_family',
             merge=salt['grains.filter_by'](osmap, grain='os',
        merge=salt['grains.filter_by'](osfingermap, grain='osfinger',
          merge=salt['pillar.get']('template:lookup', default={})
       [...]
       ---
       Traceback (most recent call last):
         File "/usr/lib/python2.7/dist-packages/salt/utils/templates.py", line 392, in render_jinja_tmpl
           output = template.render(**decoded_context)
         File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 1008, in render
           return self.environment.handle_exception(exc_info, True)
         File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 780, in handle_exception
           reraise(exc_type, exc_value, tb)
         File "<template>", line 4, in top-level template code
         File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 1073, in make_module
           return TemplateModule(self, self.new_context(vars, shared, locals))
         File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 1152, in __init__
           body_stream = list(template.root_render_func(context))
         File "/tmp/kitchen/var/cache/salt/minion/files/base/template/map.jinja", line 12, in top-level template code
           {% set defaults = salt['grains.filter_by'](default_settings,
         File "/usr/lib/python2.7/dist-packages/salt/modules/grains.py", line 600, in filter_by
           base=base)
         File "/usr/lib/python2.7/dist-packages/salt/utils/data.py", line 412, in filter_by
           for key in lookup_dict:
       TypeError: 'NoneType' object is not iterable
```

I think it should be said explicitely in the inline comments that you need to have at least an empty dict inside.

PR #94, opened 2019-04-24
fix(comments): explain that at least an empty dict is required

### LINKED ISSUE #93: or {} changes don't work for me
The changes for default empty dictionary don't work for me,...

{%- import_yaml topdir ~ "/defaults.yaml" or {} as default_settings %}
{%- import_yaml topdir ~ "/osfamilymap.yaml" or {} as osfamilymap %}
{%- import_yaml topdir ~ "/osmap.yaml" or {} as osmap %}
{%- import_yaml topdir ~ "/osfingermap.yaml" or {} as osfingermap %}

do I need a special version of a component?

myii:
* For all `os*map.yaml` files
* Remove redundant `or {}` from `map.jinja`
* Close #93

---

See main discussion in #93.  Also links back to #59 and #60.

### DIFF
--- a/template/map.jinja
+++ b/template/map.jinja
@@ -4,10 +4,10 @@
 {#- Get the `tplroot` from `tpldir` #}
 {%- set tplroot = tpldir.split('/')[0] %}
 {#- Start imports as #}
-{%- import_yaml tplroot ~ "/defaults.yaml" or {} as default_settings %}
-{%- import_yaml tplroot ~ "/osfamilymap.yaml" or {} as osfamilymap %}
-{%- import_yaml tplroot ~ "/osmap.yaml" or {} as osmap %}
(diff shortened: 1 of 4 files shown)


myii:
Thanks for the review, @n-rodriguez.

daks:
So the conclusion is that `or {}` is not needed anymore? What was the problem then? Passing the data via `context` is sufficient?

It reminds me the first version of #60 :)

myii:
@daks Here you go: https://github.com/saltstack-formulas/template-formula/issues/93#issuecomment-486099392.  I must have a very unique minion that works without the templates in place...!

myii:
> It reminds me the first version of #60 :)

You didn't fight you case hard enough there, @daks!

myii:
Thanks for the reviews, guys -- and for pointing out the error in the first place.

PR #95, opened 2019-04-24
refactor(map): use `config.get` instead of `pillar.get`

myii:
We've had some discussions in Slack/IRC/Matrix about moving on from `pillar.get` to `config.get`.  `libtofs.jinja` is already using the latter.  Selected `WIP` for this PR since this is something to consider across all formulas.  The intention here is to open up the discussion about going forward with this.

CC: @gtmanfred.

### DIFF
--- a/template/map.jinja
+++ b/template/map.jinja
@@ -14,11 +14,11 @@
     merge=salt['grains.filter_by'](osfamilymap, grain='os_family',
       merge=salt['grains.filter_by'](osmap, grain='os',
         merge=salt['grains.filter_by'](osfingermap, grain='osfinger',
-          merge=salt['pillar.get']('template:lookup', default={})
+          merge=salt['config.get']('template:lookup', default={})
         )
       )
     )
 ) %}
 
-{#- Merge the template pillar #}
-{%- set template = salt['pillar.get']('template', default=defaults, merge=True) %}
+{#- Merge the template config (e.g. from pillar) #}
+{%- set template = salt['config.get']('template', default=defaults, merge=True) %}


daks:
For now, I don't really see the use case for `config.get` instead of `pillar.get`. I read that it's to reduce load on busy masters, but I'm still not confronted to this problem.
And today all salt docs and 'litterature' (blog posts, ...) talk about pillar and none talk about anything else, like sdb. So I'm not sure if we already know what will replace pillars and there is no documentation.

In theory this has no impact because `config.get` also read pillars. But in reality I'm pretty sure that requet will take longer if all your data stays in pillars.
That's my 2 cents :)

gtmanfred:
The time difference is negligible, since it just does a lookup for the key in the `__grains__` `__opts__` or `__pillar__` object, and bails to the next one if it doesn't find it in the first.

daks:
> The time difference is negligible, since it just does a lookup for the key in the `__grains__` `__opts__` or `__pillar__` object, and bails to the next one if it doesn't find it in the first.

Good to know @gtmanfred

myii:
Thanks everyone for your feedback.  I've merged this PR.

myii:
This commit may need to be reverted.  `config.get` and `pillar.get` aren't 100% interchangeable, for example `merge=True` isn't available for the former.  So if the result of the final `config.get` is an empty dict, the entire map ends up as an empty dict.

I've tried a workaround using `defaults.merge` again but I'm running into another problem:

```salt
TypeError: Cannot update using non-dict types in dictupdate.update()
```

Anyone know of how this can be resolved?  Or is the only option to return back to `pillar.get`, for the time being at least?

myii:
Update: I've been able to find a solution to this by building upon the `defaults.merge` strategy [we used to use earlier on](https://github.com/saltstack-formulas/template-formula/blob/d7940c09988cfd9eacecf38e90872d4684eb355e/template/map.jinja).  The problem with this is that this doesn't work for `salt-ssh` users.  So the dilemma:

1. Use `config.get` via. `defaults.merge` -- get the benefit but leave `salt-ssh` behind.
1. Use `pillar.get` -- works for everyone but limited to pillar only.

PR #97, opened 2019-04-25
Doc usage

daks:


daks:
Work in progress. Not meant to be merged for now.

myii:
@daks While it would be really nice, realistically it would be better to leave off the Antora stuff for now.  Because there is another repo which controls the whole system -- I haven't shared that anywhere yet.  Actually, there's also _another_ repo, to customise the UI -- I haven't shared that, either.  I think the best idea is to work within this repo for now and then we can upgrade to Antora when the time is right.

daks:
I close this PR, and will reopen later

PR #98, opened 2019-04-26
WIP docs: adding USAGE guidelines

daks:
Work in progress. Not meant to be merged for now.
Added following a discussion with @myii and XenophonF on slack.

Replaces #97

### DIFF
--- a/docs/USAGE.rst
+++ b/docs/USAGE.rst
@@ -1,0 +1,42 @@
+.. _usage:
+
+Template formula usage
+=======================
+
+This formula implements latest 'best practices' concerning development, 
+multi-OS support, documentation and testing of SaltStack formulas.
+
+You can use it as a base to create new ones, following documentation below.
+
+You should have read https://docs.saltstack.com/en/latest/topics/development/conventions/formulas.html#writing-formulas.
+
+.. contents:: **Table of Contents**
+
+Repository clone
+----------------
+
+This repository named `template-formula` needs to be cloned in another we'll name `example-formula` in the rest of the guide. Git repository information must be removed to start from scratch your own `example` formula.
+
+```
+$ git clone https://github.com/saltstack-formulas/template-formula example-formula
+$ cd example-formula
+$ rm -fr .git
+$ git init .
+```
+
+Renaming template to example
+----------------------------
+
+With the two following commands you automatically rename all template and Template variables or functions to example and Example.
+```
+$ grep -r 'template' --exclude-dir=.git|cut -d ':' -f 1|uniq|xargs sed -i -e 's/template/example/g'
+
+# FIXME, because this affects TOFS doc too
+$ grep -r 'Template' --exclude-dir=.git|cut -d ':' -f 1|uniq|xargs sed -i -e 's/Template/Example/g'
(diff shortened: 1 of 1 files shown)


myii:
CC: @xenophonf.

myii:
Just got reminded by a question to add the notes that @xenophonf had linked to:

* https://gist.github.com/xenophonf/9cf42c0acc60d6c7241ba552daa1041d

docs/USAGE.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Renaming template to example
+----------------------------
+
+With the two following commands you automatically rename all template and Template variables or functions to example and Example.
+```
+$ grep -r 'template' --exclude-dir=.git|cut -d ':' -f 1|uniq|xargs sed -i -e 's/template/example/g'
+
+# FIXME, because this affects TOFS doc too
+$ grep -r 'Template' --exclude-dir=.git|cut -d ':' -f 1|uniq|xargs sed -i -e 's/Template/Example/g'
```
baby-gnu: I think you can do `grep -rl`, this way you don't need the `|cut -d ':' -f 1|uniq` part:

```
       -l, --files-with-matches
              Suppress normal output; instead print the  name  of  each  input
              file  from  which  output would normally have been printed.  The
              scanning will stop on the first match.
```

daks: I'm not even sure it really works, may need a more complex script.

myii: `Template` is hardly left now and doesn't need to be replaced.  This is what I've found works well, incorporating the suggestion by @baby-gnu:

```bash
$ grep -rl template . --exclude-dir=.git --exclude=.travis.yml --exclude=libtofs.jinja | xargs sed -i -e 's/template/example/g'
$ grep -rl "\- example: jinja" . --exclude-dir=.git --exclude=.travis.yml --exclude=libtofs.jinja | xargs sed -i -e 's/- example: jinja/- template: jinja/g'
```

* The first does the main replacements.
* The second fixes the use of `template` in any `file.managed` states or similar.
  - Obviously this can be improved a bit further by using back-references but it sets things up for the time being.

javierbertoli: Jumping late to the discussion, so ignore me if it does not make sense, but:

How about scripting all these commands in a file under `./bin`, like `./bin/initial_replace.sh` and just receive as a parameter the value to use? Wouldn't it be easier to maintain than documenting them?

myii: I suppose documenting the procedure was the first step towards scripting it.  However, as we appear to have reached that stage, then that's the right thing to do.  I'm sure that there will still be many guidelines that need to be documented alongside how (and when) to use this script.

daks: Yes, first I wanted to document it and wait a little bit the procedure to stabilize, before eventually writing a script.

GMAzrael: There is a problem with your search string. In root/config/file.sls, there is a file.managed state that has the "template" kv that will get picked up in the search/replace.

daks:
Closing because doc state has improved a lot in the last two years

PR #101, opened 2019-04-30
ci(kitchen+travis): use latest pre-salted images

myii:
* Add `suse` to `inspec.yml` for `opensuse-leap-15`

---

The pre-salted images have been updated, with more combinations available as well as more goodies!  This is the next step towards attaining a `5 x 3` matrix, which would have each `os` represented three times, for each of the supported salt versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,14 +20,23 @@ env:
   matrix:
     - INSTANCE: default-debian-9-2019-2-py3
     - INSTANCE: default-ubuntu-1804-2019-2-py3
-    - INSTANCE: default-centos-7-2019-2-py2
-    - INSTANCE: default-fedora-29-2019-2-py2
-    - INSTANCE: default-opensuse-423-2018-3-py2
-    - INSTANCE: default-debian-8-2018-3-py2
+    - INSTANCE: default-centos-7-2019-2-py3
+    - INSTANCE: default-fedora-29-2019-2-py3
+    - INSTANCE: default-opensuse-leap-15-2019-2-py3
+    - INSTANCE: default-debian-9-2018-3-py2
     - INSTANCE: default-ubuntu-1604-2018-3-py2
-    - INSTANCE: default-fedora-28-2018-3-py2
+    - INSTANCE: default-centos-7-2018-3-py2
+    - INSTANCE: default-fedora-29-2018-3-py2
+    # TODO: Use this when fixed instead of `opensuse-leap-42`
+    # Ref: https://github.com/netmanagers/salt-image-builder/issues/2
+    # - INSTANCE: default-opensuse-leap-15-2018-3-py2
+    - INSTANCE: default-opensuse-leap-42-2018-3-py2
     - INSTANCE: default-debian-8-2017-7-py2
     - INSTANCE: default-ubuntu-1604-2017-7-py2
+    # TODO: Enable after improving the formula to work with other than `systemd`
+    # - INSTANCE: default-centos-6-2017-7-py2
+    - INSTANCE: default-fedora-28-2017-7-py2
+    - INSTANCE: default-opensuse-leap-42-2017-7-py2
 
 script:
   - bundle exec kitchen verify ${INSTANCE}
@@ -64,4 +73,3 @@ jobs:
         script:
           # Run `semantic-release`
           - npx semantic-release@15
-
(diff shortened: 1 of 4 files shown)


myii:
@javierbertoli Updated this PR to work around the `service`-based `inspec` bug for `opensuse-leap-15`.

test/integration/default/controls/services_spec.rb
```diff
@@ -1,3 +1,7 @@
+# Temporary `if` due to `opensuse-leap-15` bug re: `service`
+if os[:name] == 'suse'
+  puts "[Skip `service`-based tests due to `opensuse-leap-15` bug]"
```
javierbertoli: Perhaps adding a reference to the bug in the comment, lile

```
Skip `service`-based tests due to `opensuse-leap-15` detection bug (see https://github.com/inspec/train/issues/377)
```
?

.travis.yml
```diff
@@ -20,14 +20,21 @@ env:
...
+    - INSTANCE: default-centos-7-2019-2-py3
+    - INSTANCE: default-fedora-29-2019-2-py3
+    - INSTANCE: default-opensuse-leap-15-2019-2-py3
+    - INSTANCE: default-debian-9-2018-3-py2
     - INSTANCE: default-ubuntu-1604-2018-3-py2
-    - INSTANCE: default-fedora-28-2018-3-py2
+    - INSTANCE: default-centos-7-2018-3-py2
+    - INSTANCE: default-fedora-29-2018-3-py2
+    # TODO: Use this when fixed instead of `opensuse-leap-42`
+    # - INSTANCE: default-opensuse-leap-15-2018-3-py2
```
javierbertoli: This can then be enabled, or should we remove it completely from here?

kitchen.yml
```diff
@@ -17,29 +17,40 @@ platforms:
...
-
+      image: netmanagers/salt-2018.3-py2:centos-7
+  - name: fedora-29-2018-3-py2
+    driver:
+      image: netmanagers/salt-2018.3-py2:fedora-29
+  # TODO: Use this when fixed instead of `opensuse-leap-42`
+  # - name: opensuse-leap-15-2018-3-py2
+  #   driver:
+  #     image: netmanagers/salt-2018.3-py2:opensuse-leap-15
+  #     run_command: /usr/lib/systemd/systemd
```
javierbertoli: This can be re-enabled?

kitchen.yml
```diff
@@ -54,11 +65,22 @@ platforms:
       image: netmanagers/salt-2017.7-py2:debian-8
   - name: ubuntu-1604-2017-7-py2
     driver:
-      image: netmanagers/salt-2017.7-py2:ubuntu-1604
-  # - name: centos-6-2017-7
+      image: netmanagers/salt-2017.7-py2:ubuntu-16.04
+  # Unavailable at the current time
+  # - name: centos-7-2017-7-py2
```
javierbertoli: This is a current distro with an old salt version. Do we want this combination? Is yes, I should add the image to the builder. If no, can you remove these comments? :yum:

test/integration/default/controls/services_spec.rb
myii: Good idea, done.

kitchen.yml
myii: Ref: https://github.com/netmanagers/salt-image-builder/pull/1.  This was a placeholder before you managed to resolve `centos-6`!  However, `centos-6` will need some effort to get working here.  So is this still valid or not?  In the meantime, I've removed the comment.

.travis.yml
myii: Thanks @javierbertoli!  I've just re-enabled the failing image.  Here are some refs for you, actually had two failures here:

* `opensuse-leap-15`: https://github.com/netmanagers/salt-image-builder/issues/2
* `opensuse-leap-42`: https://github.com/netmanagers/salt-image-builder/issues/3

Now to undo that again...

kitchen.yml
myii: See the [comment above](https://github.com/saltstack-formulas/template-formula/pull/101#discussion_r280443591).

.travis.yml
myii: Actually, that second failure was only for the `postgres-formula`, not the `template-formula`.

javierbertoli: Mind you add a link to the issue in the case of [opensuse-leap-15](https://github.com/netmanagers/salt-image-builder/issues/2) so we don't lose track of the reasons we disable things?

kitchen.yml
javierbertoli: Added an issue #104 to fix the services tests. I think this is OK for now.

myii:
Thanks, @javierbertoli.

PR #105, opened 2019-05-06
feat(centos-6): reshape formula and tests for this platform

### LINKED ISSUE #104: ci(kitchen) service tests should verify a service common across platforms
On the initial incarnations of the tests, we added `systemd-udevd` as the service to check. Now that we added `centos-6` platform, which is `SysV`, we should probably modify the test example to verify some other service, perhaps `crond`

myii:
* Close #104

---

This is an example of how #104 could be resolved, using `crond` as suggested by @javierbertoli, while making use of the `*.yaml` files as suggested by @baby-gnu.  This allows discussions to begin about how to finalise this implementation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ env:
     - INSTANCE: default-debian-8-2017-7-py2
     - INSTANCE: default-ubuntu-1604-2017-7-py2
     # TODO: Enable after improving the formula to work with other than `systemd`
-    # - INSTANCE: default-centos-6-2017-7-py2
+    - INSTANCE: default-centos-6-2017-7-py2
     - INSTANCE: default-fedora-28-2017-7-py2
     - INSTANCE: default-opensuse-leap-42-2017-7-py2
 
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -64,7 +64,7 @@ platforms:
   - name: centos-6-2017-7-py2
     driver:
       image: netmanagers/salt-2017.7-py2:centos-6
-      run_options: '--entrypoint /sbin/init'
+      run_command: /sbin/init
   - name: fedora-28-2017-7-py2
     driver:
       image: netmanagers/salt-2017.7-py2:fedora-28
--- a/pillar.example
+++ b/pillar.example
@@ -8,10 +8,16 @@ template:
   # Using bash package and udev service as an example. This allows us to
   # test the template formula itself. You should set these parameters to
   # examples that make sense in the contexto of the formula you're writing.
+  {%- if grains.osfinger == 'CentOS-6' %}
+  pkg: cronie
+  service:
+    name: crond
+  {%- else %}
   pkg: bash
-  config: /etc/template-formula.conf
   service:
     name: systemd-udevd
+  {%- endif %}
+  config: /etc/template-formula.conf
 
   tofs:
     # The files_switch key serves as a selector for alternative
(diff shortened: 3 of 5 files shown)


myii:
@daks Thanks for the review.

As a general note, I'm in two minds about leaving it with this "minimal intervention" or converting globally to `cron` instead of testing for `systemd`.

template/defaults.yaml
```diff
@@ -2,7 +2,7 @@
 # vim: ft=yaml
 ---
 template:
-  pkg: template
+  pkg: bash
```
javierbertoli: I think that the way it was before allows us to kept two clearly differentiated groups of files:

* the *formula-related* ones, which are all of those that have `template-something` or `something-template` or 'template' in the name.
* the *testing-related* ones, which are not intrinsically needed by the formulas: that's where (I think) we should put these examples, like `pkg: bash` or `service: whatever` 

I think the former group should be respected and kept free of 'testing-related' content, while the latter are the place for whatever we consider useful for testing (in fact, this is the same reason why I think that the linting and semver tools/files should be moved under `test`, at least).

If the content of the **formula-related** files is named `template` then I can simply go with a 'search & replace' for 's/template/my_package/g', the usage of a 'templated formula' is clear, and the entry-level for newcomers is easy (I usually do this, in fact :smile:).

However, if a newcomer needs to use the formula, 'search manually all around and make sense what to replace and what not, what is a requirement and what is an example', the template loses part of its value: 

'what should I replace? `pkg: bash` in the `template/defaults.yaml`? or that is a dependency that should not be modified?' 'same goes to `service: systemd-udev`?' 

'oh, I replaced `pkg: bash` but forgot of `pkg: cronie` in `template/osfingermap.yaml`. Is it a requirement or an example?'

javierbertoli:
> As a general note, I'm in two minds about leaving it with this "minimal intervention" or converting globally to cron instead of testing for systemd.

Following what I mentioned above, I think that the reason we're in this situation is because we're providing 'a testing framework' to be used with the future 'template-derived' formula but, at the same time, using the same framework to test the 'template' itself. This mean we need to keep the testing template simple while, at the same time, it has to be useful for us :yum:

I thought (not much, really :smile: ) of using `cron` because it's generally named the same across different distros. Now, in hindsight, notice that I forgot of `cronie` and `vixie-cron`, etc.

Perhaps, to keep things simple, we can use some other package and service?

myii:
@javierbertoli I can see benefits to both ways.  The points you make are definitely important.  On the flip side, it's nice having working examples of how the map runs the formula, i.e. how to specify the settings in the appropriate `.yaml` file.  If we do go with the way you've suggested, then there's a far simpler, albeit hackier, solution available.  We don't need to touch any of the `.yaml` files and just add a bit of Jinja to `pillar.example`:

```diff
+  {%- if grains.osfinger == 'CentOS-6' %}
+  pkg: cronie
+  service:
+    name: crond
+  {%- else %}
   pkg: bash
-  config: /etc/template-formula.conf
   service:
     name: systemd-udevd
+  {%- endif %}
+  config: /etc/template-formula.conf
```

And here's the working run for that:

* https://travis-ci.org/myii/template-formula/builds/530080602

This achieves the same thing as this PR.  How do you all feel about this method?

myii:
@javierbertoli As discussed on Slack, using the Jinja as mentioned in the comment above.  So this is ready to be merged.

javierbertoli:
Thanks @myii !

myii:
You're welcome, thanks for the merge, @javierbertoli.

PR #106, opened 2019-05-13
docs(semantic-release): add list of semantic-release compatible formulas

n-rodriguez:


johnkeates:
It would be nice to auto-generate such a list, but I guess this is the best we can do for now.

daks:
Not sure it's the best place to maintain this list, but it's has already been merged

myii:
@daks No problem, this can still easily be moved.  Any suggestions of a better location?

myii:
References to the discussion leading to this PR:

* https://freenode.logbot.info/saltstack-formulas/20190512#c2173801
* https://freenode.logbot.info/saltstack-formulas/20190512#c2173855
* https://freenode.logbot.info/saltstack-formulas/20190513#c2173858

daks:
not defined, could be "saltstack-formulas github project" for internal needs of contributors to start with, could also be a specific repo, or community doc.

myii:
@daks That's similar to what @n-rodriguez suggested in the linked discussions.  That's what I would also like ideally.  In fact, I was already planning to include it to the centralised Antora documentation, which I think is the ideal place for contributors and consumers alike.

PR #109, opened 2019-05-13
docs(readme): add testing section based on `postgres-formula`

myii:
If this is merged, the following task becomes active:

- [ ] Propagate this change to all `semantic-release` formulas, as listed in the documentation and #21.


myii:
Thanks for the reviews, @n-rodriguez and @johnkeates.

PR #110, opened 2019-05-13
docs(contributing): add recent `semantic-release` formulas

myii:


n-rodriguez:
actually the best thing to do would be a table with semantic-release, tests and tofs

like:

|formula|semantic-release|tofs|tests|
|---------|----------------------|-----|------|
|systemd-formula|true|true|true

myii:
@n-rodriguez I've merged this for the time being.  As we spoke about on Slack/IRC/Matrix, I've got concerns about managing a spreadsheet-like table here, at least in this documentation.  Perhaps a wiki page here?  It needs to be something easy to work with in any case.  And we need to move this section out of this file anyway, as is being worked on in #113.

PR #113, opened 2019-05-14
docs: update doc to facilitate reuse

daks:
As those docs needs to be updated when copying from this formula (cf https://github.com/saltstack-formulas/salt-formula/pull/415) I think we can simply and safely do it at the source.

**Update:** removing them is not a good idea, as they still serve to access `template-formula` own doc hosted on RTD. But to ease reuse of files on other formulas, I reorganized a bit. 

Ping https://github.com/saltstack-formulas/template-formula/pull/98


.gitignore
```diff
@@ -108,3 +108,7 @@ Gemfile.lock
 
 # copied `.md` files used for conversion to `.rst` using `m2r`
 docs/*.md
+
+# Vim
+*.swp
+*.swo
```
myii: Can be condensed into: `*.sw?`.

docs/CONTRIBUTING_DOCS.rst
```diff
@@ -0,0 +1,89 @@
+.. _contributing_docs:
+
+Contributing documentation
+==========================
+
```
myii: Let's move the `|docs|` badge here instead.

docs/README.rst
```diff
@@ -3,20 +3,20 @@
 template-formula
 ================
 
-|img_travis| |docs| |img_sr|
+|img_travis| |img_sr| |docs|
```
myii: As mentioned above, take out the `|docs|` badge from here.

docs/README.rst
```diff
@@ -3,20 +3,20 @@
...
-   :scale: 100%
-   :target: https://template-formula.readthedocs.io/en/latest/?badge=latest
 .. |img_sr| image:: https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg
    :alt: Semantic Release
    :scale: 100%
    :target: https://github.com/semantic-release/semantic-release
+.. |docs| image:: https://readthedocs.org/projects/docs/badge/?version=latest
+   :alt: Documentation Status
+   :scale: 100%
+   :target: https://template-formula.readthedocs.io/en/latest/?badge=latest
```
myii: As mentioned above, take out the `|docs|` badge from here.

myii:
@daks I've just merged #110, so that section is ready to be moved out of the file as well.  As I mentioned in the last comment there, it may be worth moving it out of the documentation and putting in the wiki instead, at least until we can find it a proper home.  What do you say, @n-rodriguez?

daks:
What do you want me to do to merge this PR?

myii:
@daks At least the review comments above.  The additional comments below (about the `semantic-release` section) can be ignored and dealt with in a separate PR.

n-rodriguez:
> What do you say, @n-rodriguez?

I agree

daks:
review comments should have been fixed

docs/README.rst
```diff
@@ -3,16 +3,12 @@
 template-formula
 ================
 
-|img_travis| |docs| |img_sr|
+|img_travis| |img_sr|
```
myii:
```suggestion
|img_travis| |img_sr|
```

myii:
Thanks, @daks!

PR #114, opened 2019-05-15
map.jinja: use tplroot

alxwr:
When updating a formula the following should be sufficient:

```
cp template-formula/template/map.jinja my-formula/my/map.jinja
sed -i '/set template =/set my =/' my-formula/my/map.jinja
```

Tested with `salt` and `salt-ssh` on FreeBSD 11.2.

@myii I'm bringing back a useful change you introduced earlier.


myii:
@alxwr `salt-ssh` clouded my mind!  Thanks for bringing this back.  I'll merge this as soon as the tests pass.

Follows on from #102.

myii:
Merged, thanks @alxwr.

PR #115, opened 2019-05-15
fix(`map.jinja`): _merge_ defaults and `config.get`

alxwr:
The merge which was formerly done by `pillar.get(..., merge=True)` has now to be performed by `grains.filter_by`.

Tested on FreeBSD with both `salt` and `salt-ssh`.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -4,6 +4,9 @@
 template:
   lookup:
     master: template-master
+    # Just for testing purposes
+    winner: lookup
+    added_in_lookup: lookup_value
 
   # Using bash package and udev service as an example. This allows us to
   # test the template formula itself. You should set these parameters to
@@ -42,3 +45,11 @@ template:
     #     - 'example_alt.tmpl'
     #     - 'example_alt.tmpl.jinja'
 
+    # For testing purposes
+    source_files:
+      template-config-file-file-managed:
+        - 'example.tmpl.jinja'
+
+  # Just for testing purposes
+  winner: pillar
+  added_in_pillar: pillar_value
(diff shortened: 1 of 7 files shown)


myii:
Follows on from #102 and #114.

Unfortunately, still triggering an error on my setup:

```
RuntimeError: maximum recursion depth exceeded while calling a Python object
```

So we've been communicating on Slack/IRC to look at an alternative solution.

alxwr:
@myii: fixed the merging, `config.get` gets only called once. Added tests to verify a correct merge.

myii:
Final solution for `salt-ssh`!

pillar.example
```diff
@@ -42,3 +45,11 @@ template:
...
     #     - 'example_alt.tmpl.jinja'
 
+    # For testing purposes
+    source_files:
+      template-config-file-file-managed:
+        - 'example.tmpl.jinja'
+
+  # Just for testing purposes
+  winner: pillar
+  added_in_pillar: asdf
```
javierbertoli: Just to make it clear that is being pulled from the correct place, wouldn't it be better to have a different value than the one `added_from_lookup` ?  ie, `added_in_pillar: pillar_value`?

template/defaults.yaml
```diff
@@ -3,6 +3,10 @@
 ---
 template:
   pkg: template
+  rootgroup: root
   config: '/etc/template'
   service:
     name: template
+  # Just here for testing
+  added_in_defaults: asdf
```
javierbertoli: same here?

pillar.example
alxwr: @javierbertoli I'll update the PR. Thanks for the feedback!

Issue #118, opened 2019-05-16
Resolve Travis concurrent jobs or consider Azure pipelines


myii:
https://freenode.logbot.info/saltstack-formulas/20190516#c2180979:

> -|-|-
> :-:|---|---
> 09:59 | myii | @​gtmanfred A problem is beginning to manifest: as we convert more formulas to `semantic-release` (and generally get `inspec` testing working on formulas), our Travis queuing is slowing things down. It only appears to handle 3 jobs at a time; the most I've ever seen is 5, but that may have been on my own forks. So it's not uncommon to see 3/50 active jobs and when things get bad, it can take up to an hour to get results.I'm usually manually cancelling and rerunning (my own) jobs to minimise loads but contributors won't be able to do that.  This load will only increase in time, as more formulas get converted.  We could start limiting the number of jobs per formula but is there anything else that can be done?
> 10:06 | slack1872 | except paying for more concurrent builds, I'm not sure...
> 10:06 | myii | I reckon we're going to have to start limiting testing per formula.
> 10:07 | myii | Unless a formula has specific tests that need to be run on specific combinations of platform/salt-version.
> 10:08 | slack1872 | Same problem always pops: you had tests _(which is great)_ and then your tests take too much time...We have the same problem @ work, with 8 kitchen VM created and highstate taking near 15 min. But on Jenkins at least we can run VM in parallel
> 10:09 | myii | What do you say to this idea:
> 10:09 | myii | 5 instances max, one from each `os`.
> 10:10 | myii | `2019.2` x2, `2018.3` x2 and `2017.7` x1.
> 10:12 | myii | We've got a nice set to _choose from_ but now this is about how to select which ones to use per formula.
> 10:12 | myii | Is 5 sufficient for testing purposes?

https://freenode.logbot.info/saltstack-formulas/20190516#c2181226:

> -|-|-
> :-:|---|---
> 12:39 | gtmanfred | no, limit concurrent jobs
> 12:39 | gtmanfred | in travis
> 12:39 | gtmanfred | you can have as many runs as you want, but it will only run one per repo at a time

https://freenode.logbot.info/saltstack-formulas/20190516#c2181237:

> -|-|-
> :-:|---|---
> 12:41 | gtmanfred | if that doesn’t work, we can look into moving to azure pipelines
> 12:41 | gtmanfred | i have been playing with that recently and it is pretty good
> 12:41 | myii | Excellent.
[...]

PR #121, opened 2019-05-19
ci(travis): reduce matrix down to 5/6 instances (ref: #118)

myii:
* The selected 5 in a "T"-shaped matrix:
```
  - 2019.2 (py3): debian-9   fedora-29   opensuse-leap-15
  - 2018.3 (py2):           ubuntu-1604
  - 2017.7 (py2):             centos6
```
* Covers each `os`, each Salt version and both Python versions

---

I was going to go for `centos-7` and `fedora-28` instead but `centos-6` is a "special" platform, which always throws up interesting situations (including not being `systemd`-based).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,17 @@ services:
 
 # Make sure the instances listed below match up with
 # the `platforms` defined in `kitchen.yml`
-# NOTE: To minimise disruption across the whole org, only select five instances
-#       to run and leave the rest commented out
-#       The recommended matrix is shown below
-#       Allow `kitchen.yml` to contain all of the platforms, to allow
-#       comprehensive local testing
+# NOTE: Please try to select up to six instances that add some meaningful
+#       testing of the formula's behaviour. If possible, try to refrain from
+#       the classical "chosing all the instances because I want to test on
+#       another/all distro/s" trap: it will just add time to the testing (see
+#       the discussion on #121).  As an example, the set chosen below covers
+#       the most used distros families, systemd and non-systemd and the latest
+#       three supported Saltstack versions with python2 and 3."
+#       As for `kitchen.yml`, that should still contain all of the platforms,
+#       to allow for comprehensive local testing
 #       Ref: https://github.com/saltstack-formulas/template-formula/issues/118
+#       Ref: https://github.com/saltstack-formulas/template-formula/issues/121
 env:
   matrix:
     - INSTANCE: default-debian-9-2019-2-py3
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,27 +13,33 @@ services:
 
 # Make sure the instances listed below match up with
 # the `platforms` defined in `kitchen.yml`
+# NOTE: To minimise disruption across the whole org, only select five instances
+#       to run and leave the rest commented out
+#       The recommended matrix is shown below
+#       Allow `kitchen.yml` to contain all of the platforms, to allow
+#       comprehensive local testing
+#       Ref: https://github.com/saltstack-formulas/template-formula/issues/118
 env:
   matrix:
     - INSTANCE: default-debian-9-2019-2-py3
-    - INSTANCE: default-ubuntu-1804-2019-2-py3
+    # - INSTANCE: default-ubuntu-1804-2019-2-py3
     - INSTANCE: default-centos-7-2019-2-py3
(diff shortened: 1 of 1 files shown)


myii:
So initial results from my fork:

Instances | Running time | Total time
:-:|--:|--:
[15](https://travis-ci.org/myii/template-formula/builds/532852168) | 10 min 8 sec | 31 min 36 sec
[5](https://travis-ci.org/myii/template-formula/builds/534541294) | 4 min 12 sec | 13 min 10 sec

myii:
Thanks for the reviews @aboe76 and @n-rodriguez.  What do you think @javierbertoli and @daks?  Can we get a merge on this so that we can start rolling out this change to the (`semantic-release`) formulas out there?

javierbertoli:
Wouldn't `centos-7` (`yum-based` installer), cover more cases (centos, redhat, fedora) than `fedora` (which installs using `dnf`)? [@aboe](https://github.com/saltstack-formulas/template-formula/issues/118#issuecomment-493707033) and [I](https://github.com/saltstack-formulas/template-formula/issues/118#issuecomment-493238121) suggested using it a few days back.

myii:
@javierbertoli BTW, you've pinged someone else there, not @aboe76!  Anyway, like I said at the top, I was going to do it using `centos-7` but I thought it would be good to have a non-`systemd` system, to cover more scenarios.  I'm relaxed either way.

javierbertoli:
> @javierbertoli BTW, you've pinged someone else there, not @aboe76! 

Ooops, fixed the typo :yum: 

> Anyway, like I said at the top, I was going to do it using `centos-7` but I thought it would be good to have a non-`systemd` system, to cover more scenarios. I'm relaxed either way.

Perhaps I was not clear, I meant replacing `fedora-29` with `centos-7`, not dropping `centos-6`.

The final matrix we suggested was more on the line:

- 2019.2 (py3): debian-9   centos-7   opensuse-leap-15
- 2018.3 (py2):           ubuntu-1604
- 2017.7 (py2):             centos-6

`centos-7` and `centos-6` differ enough as to make it reasonable to test them both. 

If the argument to add `fedora` is the same as to add `opensuse` (a RPM-based distro, different package manager), then I'd say add it to the matrix, ie, `2018.3 (py2) fedora-28`?

Perhaps I'm just nitpicking, so I'll :+1: whatever you decide.

myii:
@javierbertoli This sounds like the ideal solution, 6 in a `tree` instead of 5 in a `T`(!):

```
  - 2019.2 (py3): debian-9    centos7   opensuse-leap-15
  - 2018.3 (py2):      fedora-28   ubuntu-1604
  - 2017.7 (py2):             centos6
```

myii:
So the updated result from my fork (continuing the table above):

So initial results from my fork:

Instances | Running time | Total time
:-:|--:|--:
[15](https://travis-ci.org/myii/template-formula/builds/532852168) | 10 min 8 sec | 31 min 36 sec
[5](https://travis-ci.org/myii/template-formula/builds/534541294) | 4 min 12 sec | 13 min 10 sec
[6](https://travis-ci.org/myii/template-formula/builds/535558465) | 6 min 8 sec | 15 min 13 sec

.travis.yml
```diff
@@ -13,27 +13,33 @@ services:
 
 # Make sure the instances listed below match up with
 # the `platforms` defined in `kitchen.yml`
+# NOTE: To minimise disruption across the whole org, only select five instances
```
javierbertoli: @myii, perhaps rephrasing like this provides some insight on the reasoning to use when chosing

"...org, please try to select *up to six instances* that add some *meaningful* testing of the formula's behaviour. If possible, try to refrain from the classical "chosing all the instances because I want to test on another/all distro/s" trap: it will just add time to the testing (see the discussion on https://github.com/saltstack-formulas/template-formula/pull/121). 

As an example, the set chosen below covers the most used distros families, systemd and non-systemd and the latest three supported Saltstack versions with python2 and 3."

wdyt?

myii: Thanks, have added this comment.

javierbertoli:
Thanks @myii for your hard work, and all the rest for your feedback!

myii:
You're welcome, thank you all for your comments, reviews and feedback.  Time to start reducing our Travis burdens in the other formulas!

PR #125, opened 2019-05-24
Adding bind-formula to list of formulas with semantic release

daks:


daks:
hum, commit message is not formatted correctly, should I change it @myii ? (I may need to create a new PR)

myii:
@daks Don't worry, I'll fix it when merging.  I've got an update to make first because the `maintainer` software has been upgraded and is now breaking our builds.  I had some time earlier and merged in the changes but there's the last few configuration elements that I need to make some time for to lock down.

myii:
Fixed `maintainer` and now merged this PR -- thanks @daks.

PR #132, opened 2019-05-30
feat(MacOS): Basic package and group handling

noelmcloughlin:
This PR is a basic, but not ideal, enhancement for MacOS.

- There is no 'root' group on Darwin
- There is no `user` and `group` grains either (pity).
- The directory `/etc/template` is not created by homebrewed package installs

### DIFF
--- a/template/osfamilymap.yaml
+++ b/template/osfamilymap.yaml
@@ -10,6 +10,10 @@
 # you will need to provide at least an empty dict in this file, e.g.
 # osfamilymap: {}
 ---
+{%- if grains.os == 'MacOS' %}
+  {% set rootgroup = salt['cmd.run']("stat -f '%Sg' /dev/console") %}
+{%- endif %}
+
 Debian:
   pkg: template-debian
   config: /etc/template.d/custom.conf
@@ -40,4 +44,5 @@ Solaris: {}
 
 Windows: {}
 
-MacOS: {}
+MacOS:
+  rootgroup: {{ rootgroup | d('') }}
--- a/template/config/file.sls
+++ b/template/config/file.sls
@@ -20,6 +20,7 @@ template-config-file-file-managed:
     - mode: 644
     - user: root
     - group: {{ template.rootgroup }}
+    - makedirs: True
     - template: jinja
     - require:
       - sls: {{ sls_package_install }}


noelmcloughlin:
The template-formula has `rootgroup` only.  But the grafana-formula added `service:group` too.  We may need to add `service:group` to defaults.yaml since some requirement must exist for grafana.

template/osfamilymap.yaml
```diff
@@ -40,4 +40,9 @@ Solaris: {}
 
 Windows: {}
 
-MacOS: {}
+  {% if grains.os == 'MacOS' %}
+MacOS:
+  rootgroup: {{ salt['cmd.run']("stat -f '%Sg' /dev/console") }}
+  service:
+    group: {{ salt['cmd.run']("stat -f '%Sg' /dev/console") }}
+  {% endif %}
```
myii: So there's no standard `rootgroup` and it has to come via. that `cmd.run`?  If that's the only way, then do you mind if we tidy this up a little and remove the dupe?  How about the following suggestions:

We do the "heavy-lifting" Jinja at the top of the file:

```sls
---
{%- if grains.os == 'MacOS' %}
{%-   set macos_group = salt['cmd.run']("stat -f '%Sg' /dev/console") %}
{%- endif %}
```

And then use it below like so:

```sls
MacOS:
  rootgroup: {{ macos_group }}
  service:
    group: {{ macos_group }}
```

myii:
@noelmcloughlin Oh yes, and welcome to [`commitlint`](https://travis-ci.com/saltstack-formulas/template-formula/jobs/204370062#L502)!  It let's you know what it doesn't like about your commit message(s) -- which will end up being ignored by `semantic-release`.  So in this case, you'll want to adjust it:

```diff
-feat(MacOS): Basic package and group handling
+feat(macos): basic package and group handling
```

template/osfamilymap.yaml
noelmcloughlin: Sure that style works for me. I will try to remember to raise a salt issue asking for `sysuser` and `sysgroup` grains to be made available on MacOS as this is common issue.

Still I need to wrap usage in jinja to avoid a jinja rendering error on `macos_group`.

noelmcloughlin: Should `service:group` be added to defaults.yaml too? See grafana-formula.

myii: > Still I need to wrap usage in jinja to avoid a jinja rendering error on `macos_group`.

That's because you changed the variable names from `macos_group`!  There's a reason to avoid wrapping this section in a Jinja `if`: there's a way of simulating `map.jinja` for different OSes, but with this `if` it can't be done.  Basically, using the changes I suggested at the top of the file, you can remove the `if` and `endif` Jinja blocks now.

> Should `service:group` be added to defaults.yaml too? See grafana-formula.

Let's do it when there's a need for it.  If the need is there now, please feel free.

template/osfamilymap.yaml
```diff
@@ -10,6 +10,11 @@
 # you will need to provide at least an empty dict in this file, e.g.
 # osfamilymap: {}
 ---
+{% if grains.os == 'MacOS' %}
+  {{ set macuser = salt['cmd.run']("stat -f '%Su' /dev/console") }}
+  {{ set macgroup = salt['cmd.run']("stat -f '%Sg' /dev/console") }}
+{% endif %}
+---
```
myii: Apologies, my examples could have been clearer:

* That `---` was to specify the exact place, not to be re-included.
* Also, we're maintaining consistency of Jinja whitespace control by using `{%- ... %}` throughout this formula.
* Finally, do we need the duplicate, unused `macuser` variable at the current time?  We can add it when it becomes time to use it.

In other words, could we have this section like this instead:

```sls
{%- if grains.os == 'MacOS' %}
{%-   set macgroup = salt['cmd.run']("stat -f '%Sg' /dev/console") %}
{%- endif %}
```

myii: And if `macuser` is required due to the situation mentioned below, then there's no need to duplicate the `cmd.run`.  Simply use the `macgroup` variable.  Or vice versa.

myii:
@noelmcloughlin Don't worry, the `commitlint` takes a little getting used to but you can end up appreciating the standardised format in the end.  And the structured Changelog, etc. are well worth it.

template/osfamilymap.yaml
noelmcloughlin: thanks for review @myii   You are correct.  I have updated the commit accordingly.

So on non-MacOS the `rootgroup` variable (at bottom of osfamilymap.yaml) will resolve?  I was'nt aware of that.

template/osfamilymap.yaml
noelmcloughlin: Sure we can leave it out. 
If `macuser` (or more generically `sysuser`) is needed we can add this.  The second `cmd.run` would be needed as command to detect `user` and `group` differs.

noelmcloughlin: https://github.com/saltstack/salt/issues/53325

myii:
Merged, thanks for the patience @noelmcloughlin.  Now that we're using the `rootgroup` variable, I might reshape things a little, to prevent confusion with `*BSD`.  That's for another PR, though.

noelmcloughlin:
Thanks @myii  I raised salt enhancement request  https://github.com/saltstack/salt/issues/53325

myii:
> ... I raised salt enhancement request [saltstack/salt#53325](https://github.com/saltstack/salt/issues/53325)

@noelmcloughlin Thanks for doing that.

PR #133, opened 2019-05-31
refactor(`osfamilymap`): avoid *BSD ambiguity with MacOS `rootgroup`

myii:
Follows-on from #132.

### DIFF
--- a/template/osfamilymap.yaml
+++ b/template/osfamilymap.yaml
@@ -11,7 +11,7 @@
 # osfamilymap: {}
 ---
 {%- if grains.os == 'MacOS' %}
-  {% set rootgroup = salt['cmd.run']("stat -f '%Sg' /dev/console") %}
+{%-   set macos_rootgroup = salt['cmd.run']("stat -f '%Sg' /dev/console") %}
 {%- endif %}
 
 Debian:
@@ -45,4 +45,4 @@ Solaris: {}
 Windows: {}
 
 MacOS:
-  rootgroup: {{ rootgroup | d('') }}
+  rootgroup: {{ macos_rootgroup | d('') }}


noelmcloughlin:
This PR is fine - the jinja variable applies to MacOS.   But I cannot find any official documentation linking the term `rootgroup` or `root group`  to either Linux or BSD. 

It's possible to find explanations like this ... but not the term `rootgroup`?
> If you are provisioning a FreeBSD or OpenBSD host, then the group for the root will be wheel, but on an Ubuntu machine, the group is named admin. [https://books.google.ie/books?isbn=1788399382]

Could you add a comment to defaults.yaml clarifying what variable is modeling?

myii:
> ... But I cannot find any official documentation linking the term `rootgroup` or `root group` to either Linux or BSD.
> ...
> Could you add a comment to defaults.yaml clarifying what `rootgroup` variable is modeling?

@noelmcloughlin Looking at a global search through the org (https://github.com/search?q=org%3Asaltstack-formulas+rootgroup&type=Code), you'll find that `rootgroup` has been the exclusive `.yaml` favourite of one of our regular contributors (and `*BSD` / `salt-ssh` advocate): @alxwr!    Obviously, always feeling lonely with his sorry `wheel`, he has decided to force `rootgroup` upon us all!

It's always the same usage in the formulas themselves:

```sls
    - group: {{ <tplroot>.rootgroup }}
```

Basically, it's just a YAML value that looks like it's becoming a SaltStack Formulas convention for dealing with non-`root` root group's, such as `*BSD`.  I'll let @alxwr respond with any PR to add that clarification.  Thanks for bringing this up @noelmcloughlin!

noelmcloughlin:
I don't mind using term `rootgroup` but I am curious whether term has special meaning. If the meaning is ambiguous this will bite us later so thats why I was suggesting adding a comment to clarify our position on why we use the definition

myii:
@noelmcloughlin I think it was just an obvious term to use in the `.yaml` files when trying to find a way to describe needing to use `wheel` as the root group instead of `root`.  There's no special significance to it otherwise.  It would be worth mentioning that, alongside the notes that `MacOS` doesn't have a root group and so needs the Jinja to get us a certain value.

By the way, I was meaning to ask you: can you share some examples of what the command `stat -f '%Sg' /dev/console` returns on `MacOS`?  I'm interested to see if there are any patterns between different minions.  Or at least understand why we have to resort to this hack in the first place.

myii:
@noelmcloughlin Fantastic, who needs the Jinja?  We can simply use:

```sls
MacOS:
  rootgroup: nmcloughlin
```

for `MacOS`!

I think the general breakdown across SaltStack Formulas can be summarised as:

OS | `rootgroup`
---|---
Linux | `root`
BSD | `wheel`
MacOS | `nmcloughlin`(!)
Windows | `???`

Is it `Administrator` and/or `System` in Windows (or does it even apply in any way)?  And there may be exceptions here or there but most commonly (for formulas), `rootgroup` == `root`.

noelmcloughlin:
Windows rootgroup is `Administrator` group - populated by users.
Ubuntu rootgroup is `admin` not `wheel` but I think `root` works fine?

myii:
@noelmcloughlin I've been using Ubuntu for a long time now.  I've never come across the `admin` group; it's always been `root`.

noelmcloughlin:
I know `admin` is related to sudoers on Debian; On RedHat you use `wheel` I think.  But thats different discussion. 

Salt has poor group management in general and hard to get group management fixed: https://github.com/saltstack/salt/issues/46361

myii:
> I know `admin` is related to sudoers on Debian; On RedHat you use `wheel` I think. But thats different discussion.

@noelmcloughlin If you do find something conclusive, it would be good to add to the [`sudoers-formula` map](https://github.com/saltstack-formulas/sudoers-formula/blob/1a46975212b5e0c6c8162c8b15982ae4966933b6/sudoers/map.jinja).  The only `wheel` there so far is for `FreeBSD`; and no `admin`.

PR #131, opened 2019-05-27
chore(`libsaltcli`): add lib to check type of Salt command being used

myii:
* To distinguish between:
  - `salt-minion`
  - `salt-call`
  - `salt-ssh`
* Invoked like `map.jinja`:
  - `{%- from tplroot ~ "/libsaltcli.jinja" import cli with context %}`
* Based upon work done in PRs: #102, #114 & #115

Issue #122, opened 2019-05-20
[Discussion] [Idea] Move all utilities for formulas outside of edited files (states, yaml)


daks:
On several discussions about TOFS or config.get, I have expressed my main problem with those 'tools': they clutter/complicate the files that contributors needs to write.

My point of view on this is that such basic/simple formulas tools should not complicate any new comer which want to create a simple formula without any knowledge of them before. 
They are too intrusive to keep this principe.

Part of TOFS is already abstracted in `libtofs.jinja` but it still needs 4-5 lines for each `file.managed`.
I know less config.get but I think it also had a lot of code _(can't find now the PR I am looking)_ and especialy to deal with cases like salt-ssh.

I wonder if there is some way to do it. Puppet modules uses `lib` to add specific code. Ansible roles seems to use `handlers`.
Couldn't `_modules` or `_states` or whatever be used to do it? Even if we need to convert jinja/state code to Python.

cc @myii, @javierbertoli @aboe76 @n-rodriguez to open-mindly discuss this idea :)

aboe76:
@daks I think you have point there, `_modules` should be picked up by salt if you put them in the formula.

But you have more those are called `extension_mdules`
https://docs.saltstack.com/en/latest/ref/configuration/master.html#extension-modules

and https://github.com/saltstack/salt-contrib has a bunch of them

aboe76:
Or get them to add libtofs, into the core state system.

alxwr:
@daks Thanks for posting an alternative approach to #116!

> I know less config.get but I think it also had a lot of code _(can't find now the PR I am looking)_ and especialy to deal with cases like salt-ssh.

Maybe #102 and #115?

https://github.com/saltstack-formulas/template-formula/blob/91bc2f046b3629b9f1f66ed0eacddb626348da6c/template/map.jinja

This is roughly the same amount of code which was needed before with `pillar.get`. @myii and I just used more lines to increase readability. Really the main difference this: `pillar.get` got replaced by `config.get` and all merging is now done by `grains.filter_by`.
The recent fuzz about `salt-ssh` comes from minor implementation differences between it and regular `salt` (and my stumbling over them). The merging (which makes up most of the code) has to be done for `salt` too.

@daks I wholeheartedly agree with your point, that using `_modules` and/or `_states` is a nicer approach. I.e. `states.file.tofs`¹ and `configstack.get`² would be much cleaner and make a formula much more maintainable than the current way.

[...]

PR #134, opened 2019-06-01
feat(jinja): add macros.jinja file to accumulate common macros

noelmcloughlin:
This PR introduces a 1st revision of a common `jinja/macros.j2` file to gather generic macros.  The first revision includes a `format_kwargs` macro.

**Background:**
The `format_kwargs` macro is generic solution used in two formulas-
- ceph-formula
- postgres-formula

**format_kwargs macro**
This macro makes short work of rendering SLS lists which must be generated by mapping files and the switching of `grains`.

**Use Case**

osfamilymap.yaml
```
Debian:
  pkg:
    repo:
      name: 'deb https://packages.template.org/debian stable main'
      key_url: https://packages.template.org/debian/template.org.key
      file: file: /etc/apt/sources.list.d/template.list
      dist: stable

RedHat:
  pkg:
    repo:
      name: template-cli
      baseurl: https://packages.template.org/redhat
      key_url: https://packages.template.org/redhat/template.org.key
      enabledTrue: 1
```

package/install.sls
```
{%- from tplroot ~ "/jinja/macros.j2" import format_kwargs with context -%}

template-package-repo-managed:
  pkgrepo.managed:
    {{- format_kwargs(template.pkg.repo) }}
    - require_in:
      - pkg: template-package-install-pkg-installed
```

### DIFF
--- a/template/jinja/macros.jinja
+++ b/template/jinja/macros.jinja
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# vim: ft=jinja
+#
+# Collection of common macros
+
+{%- macro format_kwargs(kwarg) %}
+
+  {%- filter indent(4) %}
+    {%- for k, v in kwarg|dictsort() %}
+- {{ k }}: {{ v }}
+    {%- endfor %}
+  {%- endfilter %}
+
+{%- endmacro %}


template/jinja/macros.j2
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# vim: ft=jinja
+#
+# Collection of common macros
+
+{%- macro format_kwargs(kwarg) -%}
```
myii: Need to be cautious with Jinja whitespace control `{%- ... -%}`: https://github.com/saltstack-formulas/collectd-formula/pull/141#issuecomment-485043263.

myii:
Thanks for this, @noelmcloughlin.  I think the idea itself is sound but we all need discuss how to proceed, with respect to consistency and a number of ongoing discussions:

1. Using a `jinja` sub-directory; we haven't done so far, such as with `libtofs.jinja` but it's probably a good idea to separate these libraries _somewhere_.
1. Using the `.j2` file extension; we're primarily using `.jinja` so far; when's the right time to use `.j2` and the right time to use `.jinja`?
1. Should we group numerous macros under `macros.xyz` or should we have separate libraries for each one, with consideration to spreading updates across numerous formulas?
1. How should we implement this with regards to #122?
1. Any decisions here will also affect #131.

If we're going to follow our current implementation scheme, then this should be something like:

```jinja
{%- from tplroot ~ "/libkwarg.jinja" import format_kwargs with context -%}
```

However, I am _not_ proposing that we continue with this scheme.

noelmcloughlin:
Hi @myii 

> Using a jinja sub-directory; we haven't done so far, such as with libtofs.jinja but it's probably a good idea to separate these libraries somewhere.

macros are an advanced feature of a rendering language. It is probably a sound principle to hide the implementation since they are secondary libraries designed to cleanup overuse of jinja in SLS files.

> when's the right time to use .j2 and the right time to use .jinja? Is it worth using one for libraries and the other for templates?

Fixed. Let's stick with `jinja` because the version is just an implementation detail.

> Should we group numerous macros under macros.xyz or should we have separate libraries for each one, with consideration to spreading updates across numerous formulas?

There is no right answer.  It's unclear if loading a single macro from a common library file incurs increased performance, compared to a modular library implementation. Buts It's a more simple implementation. We could start here and evolve.

This can wait until #122 and #131 become clearer.

noelmcloughlin:
Here is another macro from prometheus-formula (only) which maybe portable too.
```
{#- Contactenate arguments #}
{%- macro concat_args(args) %}
{%-   set args = args|dictsort %}
{%-   if args|length > 0 %}
{%-     for k,v in args -%}
{%-       if not k or not v %}{% continue %}{% endif -%}
          --{{ k }}={{ v }}
{%-       if not loop.last %} {% endif -%}
{%-     endfor -%}
{%-   endif -%}
{%- endmacro %}
```
Anyway wait until #122 and #131 become clearer.

noelmcloughlin:
I'll close this as duplicate or #122

PR #136, opened 2019-06-04
docs(readme): show only one level in table of contents

noelmcloughlin:
This PR makes the table of contents less verbose (depth: 1).

```
Table of Contents

General notes
Contributing to this repo
Available states
Testing
```

The default depth (two) is too verbose for users.  

The list of available states is not impacted by this update.

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -18,6 +18,7 @@ A SaltStack formula that is empty. It has dummy content to help with a quick
 start on a new formula and it serves as a style guide.
 
 .. contents:: **Table of Contents**
+   :depth: 1
 
 General notes
 -------------


noelmcloughlin:
OpenSuse Travis CI job failed. `bundler: failed to load command: kitchen (/home/travis/build/saltstack-formulas/template-formula/vendor/bundle/ruby/2.4.0/bin/kitchen)`

myii:
@noelmcloughlin I can see advantages to both ways, so I'm not convinced on proceeding with the merge.  The main reason being is that the current method has been in use for a long time (i.e. before all of these recent developments) so using this method would be inconsistent until it eventually propagated through all of the repos out there.  The other thing is that we really should have the proper documentation solution in place before this propagation can complete, so that will introduce even more inconsistencies in the interim.  Essentially, it may end up being a change for no real gain.  However, I'm not going to get in the way if this is preferred by others as well.

noelmcloughlin:
The repetition of "Available States" from usage of both  **".. contents:: table of contents"**and **"contents:: local"** is unnecessary and removing the repetition feels more user friendly.  An alternative option is to remove `.. contents:: table of content` from the README.  However since documentation is WIP I can close this PR. wdyt

myii:
@noelmcloughlin No need to close it, this is a valid idea.  It's just horses for courses.  Personally, I've got very used to seeing everything at the top, even since dealing with SaltStack Formulas.  So I'd suppose that there are a few others like me, who wouldn't like to have to scroll down to "find" the states' list.  Yes, there's some duplication but it doesn't cause any harm from an RST perspective.  Still, some may suggest we should have one or the other, not both.  And then there are some who prefer both...!

noelmcloughlin:
It's hard to RTFM on some formula with **many states**.  Lots of scrolling with mouse.

myii:
@daks @n-rodriguez Can I pull you both in for this PR and number #169?  You've both done a number of `semantic-release` conversions.  Your opinions would be very useful, since this will lead to changing the structure across all of the formulas.

Personally, I'm OK either way.  My only concern is _consistency_; whatever we agree on should be used across the org.

n-rodriguez:
> I approve this PR since I find it more readable for new users

#metoo

noelmcloughlin:
I'll handle  #169 in another PR because this is now an "unknown branch". /shrug
Thanks, everyone for reviewing. Appreciated.

PR #146, opened 2019-06-25
logrotate-formula is compatible semantic-release

daks:


myii:
@daks Had to fix the commit message but this has been merged, thanks.

PR #148, opened 2019-07-09
ci(kitchen+travis): replace EOL pre-salted images

myii:
```
* Fedora:   30 => 29 & 29 => 28 (EOL)
* OpenSuse: 15 => 42 (EOL)
  - Use `dist: xenial` to prevent failures
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,26 +30,30 @@ services:
 #       Ref: https://github.com/saltstack-formulas/template-formula/issues/121
 env:
   matrix:
-    - INSTANCE: default-debian-9-develop-py3
+    - INSTANCE: default-debian-10-develop-py3
     # - INSTANCE: default-ubuntu-1804-develop-py3
     # - INSTANCE: default-centos-7-develop-py3
-    # - INSTANCE: default-fedora-29-develop-py3
+    # - INSTANCE: default-fedora-30-develop-py3
     # - INSTANCE: default-opensuse-leap-15-develop-py3
+    # - INSTANCE: default-amazonlinux-2-develop-py2
     # - INSTANCE: default-debian-9-2019-2-py3
     - INSTANCE: default-ubuntu-1804-2019-2-py3
-    - INSTANCE: default-centos-7-2019-2-py3
-    # - INSTANCE: default-fedora-29-2019-2-py3
+    # - INSTANCE: default-centos-7-2019-2-py3
+    # - INSTANCE: default-fedora-30-2019-2-py3
     # - INSTANCE: default-opensuse-leap-15-2019-2-py3
+    - INSTANCE: default-amazonlinux-2-2019-2-py2
     # - INSTANCE: default-debian-9-2018-3-py2
     # - INSTANCE: default-ubuntu-1604-2018-3-py2
     # - INSTANCE: default-centos-7-2018-3-py2
     - INSTANCE: default-fedora-29-2018-3-py2
-    - INSTANCE: default-opensuse-leap-42-2018-3-py2
+    - INSTANCE: default-opensuse-leap-15-2018-3-py2
+    # - INSTANCE: default-amazonlinux-2-2018-3-py2
     # - INSTANCE: default-debian-8-2017-7-py2
     # - INSTANCE: default-ubuntu-1604-2017-7-py2
     - INSTANCE: default-centos-6-2017-7-py2
-    # - INSTANCE: default-fedora-28-2017-7-py2
-    # - INSTANCE: default-opensuse-leap-42-2017-7-py2
+    # - INSTANCE: default-fedora-29-2017-7-py2
+    # - INSTANCE: default-opensuse-leap-15-2017-7-py2
+    # - INSTANCE: default-amazonlinux-2-2017-7-py2
 
 script:
   - bin/kitchen verify ${INSTANCE}
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -12,9 +12,9 @@ driver:
 # the `env.matrix` instances defined in `.travis.yml`
 platforms:
   ## SALT `develop`
-  - name: debian-9-develop-py3
+  - name: debian-10-develop-py3
     driver:
-      image: netmanagers/salt-develop-py3:debian-9
+      image: netmanagers/salt-develop-py3:debian-10
       provision_command:
         - curl -o bootstrap-salt.sh -L https://bootstrap.saltstack.com
         - sh bootstrap-salt.sh -XdPbfrq -x python3 git develop
@@ -30,9 +30,9 @@ platforms:
       provision_command:
         - curl -o bootstrap-salt.sh -L https://bootstrap.saltstack.com
         - sh bootstrap-salt.sh -XdPbfrq -x python3 git develop
-  - name: fedora-29-develop-py3
+  - name: fedora-30-develop-py3
     driver:
-      image: netmanagers/salt-develop-py3:fedora-29
+      image: netmanagers/salt-develop-py3:fedora-30
       provision_command:
         - curl -o bootstrap-salt.sh -L https://bootstrap.saltstack.com
         - sh bootstrap-salt.sh -XdPbfrq -x python3 git develop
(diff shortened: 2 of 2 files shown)


myii:
Updated to include the new `amazonlinux-2` images as well as `debian-10` (`develop` only for the time being).

kitchen.yml
```diff
@@ -79,6 +88,9 @@ platforms:
     driver:
       image: netmanagers/salt-2018.3-py2:opensuse-leap-15
       run_command: /usr/lib/systemd/systemd
+  - name: amazonlinux-2-2018-3-py2
+    driver:
+      image: netmanagers/salt-2018.3-py2:amazonlinux-2
 
   ## SALT 2017.7
   - name: debian-8-2017-7-py2
```
daks: Same for here.

.travis.yml
```diff
@@ -26,30 +27,30 @@ services:
...
     # - INSTANCE: default-ubuntu-1604-2018-3-py2
     # - INSTANCE: default-centos-7-2018-3-py2
     - INSTANCE: default-fedora-29-2018-3-py2
-    # TODO: Use this when fixed instead of `opensuse-leap-42`
-    # Ref: https://github.com/netmanagers/salt-image-builder/issues/2
-    # - INSTANCE: default-opensuse-leap-15-2018-3-py2
-    - INSTANCE: default-opensuse-leap-42-2018-3-py2
+    - INSTANCE: default-opensuse-leap-15-2018-3-py2
+    # - INSTANCE: default-amazonlinux-2-2018-3-py2
     # - INSTANCE: default-debian-8-2017-7-py2
```
daks: You changed previous debian-9 to debian-10. I would have changed this debian-8 commented line instead.

Maybe also let debian-9 be tested and the new debian-10 recently released be tested too, keeping both for some time.

myii: @daks Thanks for the suggestion.  With so many images to choose from, it's always going to be hard to select 6 for the matrix!  The general idea behind the Debian images was to have all of the supported versions represented at least once.  So that became:

* `10` x 1
* `9` x 2
* `8` x 1

In terms of the matrix itself, we've now got 6 different platforms, so we need to use each one once.  Since `amazonlinux-2` was a straight swap with `centos-7`, that's what I did.

Ultimately, each formula can use it's own combination, so I'm not sure how to improve this without losing something else.

daks: I would say that, even if it's still supported one year, Debian 8 is now the least critical to remove for Debian versions.

aboe76: It might be easier to track the debian version from the saltstack repo, debian 10 packages are still a wip.

myii: @javierbertoli I'd love to get your input on this discussion above.

myii: Some more information about this.

https://saltstackcommunity.slack.com/archives/C8E7RJCPN/p1562940070040900:

> David Murphy
> The Debain 10 packages should appear in the next point release which is Salt 2019.2.1 and which is nearing the end of its testing in QA
> David Murphy
> FYI - https://github.com/saltstack/salt-pack-py3/issues/117#issuecomment-510872352 for those who cannot wait, note not official SaltStack build

* There's some conversation above and below this comment as well.

javierbertoli: I kinda agree with @daks, that we should keep them for a while, but also remember we don't need to use/enable them all as long as the *relevant* tests so a matrix:

* latest os-deb, py3, git salt
* latest os-rpm, py3, git salt
* latest os-deb, py3, packaged salt
* latest os-rpm, py3, packaged salt
* previous os-deb, py2, packaged salt
* previous os-rpm, py2, packaged salt
* oldest os-deb, py2, packaged salt
* oldest os-rpm, py2, packaged salt

would cover most of the relevant cases. I think that:

```yaml
# latest os-deb, py3, git salt
- INSTANCE: default-debian-10-develop-py3
# - INSTANCE: default-ubuntu-1804-develop-py3
# - INSTANCE: default-centos-7-develop-py3

# latest os-rpm, py3, git salt
- INSTANCE: default-fedora-30-develop-py3
# - INSTANCE: default-opensuse-leap-15-develop-py3
# - INSTANCE: default-amazonlinux-2-develop-py2
# - INSTANCE: default-debian-9-2019-2-py3 

# latest os-deb, py3, packaged salt
- INSTANCE: default-ubuntu-1804-2019-2-py3

# latest os-rpm, py3, packaged salt
- INSTANCE: default-centos-7-2019-2-py3
# - INSTANCE: default-fedora-30-2019-2-py3
# - INSTANCE: default-opensuse-leap-15-2019-2-py3
# - INSTANCE: default-amazonlinux-2-2019-2-py2  

# previous os-deb, py2, packaged salt
# - INSTANCE: default-debian-9-2018-3-py2
# - INSTANCE: default-ubuntu-1604-2018-3-py2            
# - INSTANCE: default-centos-7-2018-3-py2                   
# - INSTANCE: default-fedora-29-2018-3-py2                      

# previous os-rpm, py2, packaged salt
- INSTANCE: default-opensuse-leap-15-2018-3-py2                     
# - INSTANCE: default-amazonlinux-2-2018-3-py2                          

# oldest os-deb, py2, packaged salt
- INSTANCE: default-debian-8-2017-7-py2
# - INSTANCE: default-ubuntu-1604-2017-7-py2                                    

# oldest os-rpm, py2, packaged salt
- INSTANCE: default-centos-6-2017-7-py2
# - INSTANCE: default-fedora-29-2017-7-py2                                              
# - INSTANCE: default-opensuse-leap-15-2017-7-py2                                           
# - INSTANCE: default-amazonlinux-2-2017-7-py2    
```

might satisfy those cases?

aboe76: @javierbertoli that would cover most OSes and derivatives

myii: @daks @aboe76 @javierbertoli Can we reach a decision on this soon?  The first stage of PRs to standardise the files has almost completed (thanks to you all for your reviews and merges).  The second stage will depend on this decision; I've also got a `users-formula` PR waiting in the wings.

Actually, there is another way: we could wait for me to get `ssf-formula` out there and then have a PR there to make this final decision.  Perhaps that would be even better, since all the PRs would be triggered from a PR merge in that formula...

So another question: I've been considering where to put this `ssf-formula` repo.  I don't want to be presumptuous and put it into this org immediately.  I was thinking to start the repo in my account and test it out for a few more iterations.  Then we could decide to bring it across at the right time.  What do you think?

daks: @myii looks fine to put this tool in your personal account (that's where I looked for it :) at least for now. Then we could move it to org.

About the images list, the one from @javierbertoli looks fine.

myii: @daks Thanks for the feedback, I'll start my repo in my account, as planned.

In terms of the matrix, a few queries:

* How about representing the new `amazonlinux` images we have?
* I thought we agreed to limit ourselves to no more than 6 images per formula -- as I pushed through all of the PRs with `ssf-formula`, there was a logjam for an hour or two.
* We have an exception on the `vault-formula`, where we're keeping hold of a platform that still uses `upstart` -- is it worth ensuring we have one of these in the matrix as well, such as an Ubuntu-14.04 (which is still under support for a while yet)?

myii:
Rebased and squashed after merging #153.

aboe76:
@myii shall I merge this?

myii:
@aboe76 I'm in no hurry now since merging #153.  I'm going to share the `ssf-formula` soon and it can be discussed there, since we can compare the effects on all of the other formulas as well.

myii:
@aboe76 Actually, would you mind merging this?  Then it will match the current state of `ssf-formula` and the new `semantic-release` PRs that I've prepared.

myii:
We can always resolve this via. `ssf-formula` afterwards, easily.

myii:
Thanks @aboe76 for the merge.  I'll ping you all for the discussion about (propagating) this using the `ssf-formula`.  Then we can finalise the matrix as required.

myii:
The plan is to use https://github.com/myii/ssf-formula/pull/16 to propagate the changes to all of the repos.

PR #150, opened 2019-07-23
Bugfix/libtofs crash on files switch lookup list

baby-gnu:


baby-gnu:
This results in the following output on my test machine:

```
       template-config-file-file-managed:
         file.managed:
           - name: /etc/template-formula.conf
           - source: 
             - salt://template/files/any/path/can/be/used/here/example.tmpl.jinja
             - salt://template/files/e4585afd55d3/example.tmpl.jinja
             - salt://template/files/Debian-9/example.tmpl.jinja
             - salt://template/files/foo/example.tmpl.jinja
             - salt://template/files/bar/example.tmpl.jinja
```

The two lines before the `default` come from the `role`.

I'm not really happy with putting the `role` directly in `pillar.example`, shouldn't we have a dedicated pillar for tests? Maybe even using both of them?

baby-gnu:
> I'm not really happy with putting the `role` directly in `pillar.example`, shouldn't we have a dedicated pillar for tests? Maybe even using both of them?

Maybe something like

```diff
diff --git a/kitchen.yml b/kitchen.yml
index 3be550a..d62989d 100644
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -124,8 +124,10 @@ provisioner:
       base:
         '*':
           - template
+          - test
   pillars_from_files:
     template.sls: pillar.example
+    test.sls: test/salt/pillar/test.sls
 
 verifier:
   # https://www.inspec.io/
diff --git a/pillar.example b/pillar.example
index fd8a5b8..1b035ec 100644
--- a/pillar.example
+++ b/pillar.example
@@ -56,7 +56,3 @@ template:
   # Just for testing purposes
   winner: pillar
   added_in_pillar: pillar_value
-
-role:
-  - foo
-  - bar
diff --git a/test/salt/pillar/test.sls b/test/salt/pillar/test.sls
new file mode 100644
index 0000000..5ac428c
--- /dev/null
+++ b/test/salt/pillar/test.sls
@@ -0,0 +1,4 @@
+# libtofs.jinja must work with tofs.files_switch looked up list
+role:
+  - foo
+  - bar
```

myii:
@baby-gnu That's a prompt response!  Thanks for running with this, I'll check it out as soon as I get a chance.

@ErisDS You may be interested in this, since this was instigated by our discussion on Slack yesterday.

myii:
@baby-gnu Using an extra pillar for testing is definitely a solution we can work with.  I've already adapted `ssf-formula` to deal with this, so fine at my end.

baby-gnu:
> Using an extra pillar for testing is definitely a solution we can work with. I've already adapted `ssf-formula` to deal with this, so fine at my end.

Thanks @myii, I'll integrate this.

Where is the `ssf-formula` to have a look?

myii:
@baby-gnu I'll be sharing it soon!  See https://github.com/saltstack-formulas/cert-formula/pull/28#issuecomment-513866974 for a bit more explanation.

template/libtofs.jinja
```diff
@@ -83,18 +83,26 @@
...
+          {%- set fs_dir_lookup = salt['config.get'](tplroot ~ ':tofs:dirs:default', 'default') %}
         {%- endif %}
-        {%- set url = [
-            '- salt:/',
-            path_prefix_inc_ext.strip('/'),
-            files_dir.strip('/'),
-            fs_dir.strip('/'),
-            src_file.strip('/'),
-        ] | select | join('/') %}
+	{#- Force the `config.get` lookup result as a list #}
```
myii:
```suggestion
+	{#- Force the `config.get` lookup result as a list where necessary #}
+	{#- since we need to also handle grains that are lists #}
```

template/libtofs.jinja
```diff
@@ -83,18 +83,26 @@
...
+	{%- if fs_dir_lookup is string %}
+	  {%- set fs_dirs = [fs_dir_lookup] %}
+	{%- else %}
+	  {%- set fs_dirs = fs_dir_lookup %}
+	{%- endif %}
+	{%- for fs_dir in fs_dirs %}
```
myii: We can lose the `else` and the extra variable doing it this way instead:

```diff
-	  {%- set fs_dirs = [fs_dir_lookup] %}
-	{%- else %}
-	  {%- set fs_dirs = fs_dir_lookup %}
+	  {%- set fs_dir_lookup = [fs_dir_lookup] %}
 	{%- endif %}
-	{%- for fs_dir in fs_dirs %}
+	{%- for fs_dir in fs_dir_lookup %}
```

myii:
@baby-gnu By the way, I forgot to mention that I tested my suggested changes and all is looking good:

* https://travis-ci.org/myii/template-formula/builds/562558590

Furthermore, if you do agree to change to `roles` instead, I sorry for having to get you to modify both/all of your commits!

template/libtofs.jinja
baby-gnu: I apply what you suggested but used the `fs_dirs` for the lookup result.

baby-gnu:
I just fixed the commit messages `role` → `roles`.

Thanks @myii for the note in IRC.

myii:
Merged, thanks for the important contribution @baby-gnu!

PR #152, opened 2019-07-23
fix(tofs): prepend the config-based `source_files` to the default

### LINKED ISSUE #151: Change file_switch to append source_file lookups, rather than override
To give a concrete example from [nginx-formula](https://github.com/TryGhost/nginx-formula/blob/e957a022763db369dece479479301edaa7223c91/nginx/servers_config.sls#L116
):

```
 {{ files_switch([server, 'server.conf'],
                      'server_conf_file_managed'
         )
      }}
```

On any given host, I may have many managed `server` values. Therefore, it's important that it's possible to use the dynamic `server` filename.

If I configure `nginx.tofs.source_files.server_conf_file_managed` to, e.g. `default.conf`, I expect this to be appended to the list of files that get looked for.

However instead, nginx-formula will now only look for `default.conf`, removing the ability to have multiple servers per host. 

My expectation is that setting server_conf_file_managed adds a cascading override - i.e. it appends to the beginning of the list of file names to look for, not overrides the full existing list.

myii:
* https://github.com/saltstack-formulas/nginx-formula/pull/247#issuecomment-514262549
  - The main issue is that the `nginx-formula` has dynamic values being used as the default `source_files` -- there is no way to provide this from the pillar/config in a sensible fashion
  - Prepending to this default (rather than overriding it) resolves this problem entirely, without adding excessive entries to the `source`
* Closes #151

---

CC: @baby-gnu.

### DIFF
--- a/docs/TOFS_pattern.rst
+++ b/docs/TOFS_pattern.rst
@@ -325,6 +325,7 @@ We can simplify the ``conf.sls`` with the new ``files_switch`` macro to use in t
 
 
 * This uses ``config.get``, searching for ``ntp:tofs:source_files:Configure NTP`` to determine the list of template files to use.
+* If this returns a result, the default of ``['/etc/ntp.conf.jinja']`` will be appended to it.
 * If this does not yield any results, the default of ``['/etc/ntp.conf.jinja']`` will be used.
 
 In ``libtofs.jinja``, we define this new macro ``files_switch``.
@@ -426,7 +427,6 @@ The list of ``source_files`` can be given:
      tofs:
        source_files:
          Configure NTP:
-           - '/etc/ntp.conf.jinja'
            - '/etc/ntp.conf_alt.jinja'
 
 Resulting in:
@@ -434,10 +434,13 @@ Resulting in:
 .. code-block:: sls
 
          - source:
-           - salt://ntp/files/theminion/etc/ntp.conf.jinja
            - salt://ntp/files/theminion/etc/ntp.conf_alt.jinja
-           - salt://ntp/files/Debian/etc/ntp.conf.jinja
+           - salt://ntp/files/theminion/etc/ntp.conf.jinja
            - salt://ntp/files/Debian/etc/ntp.conf_alt.jinja
-           - salt://ntp/files/default/etc/ntp.conf.jinja
+           - salt://ntp/files/Debian/etc/ntp.conf.jinja
            - salt://ntp/files/default/etc/ntp.conf_alt.jinja
+           - salt://ntp/files/default/etc/ntp.conf.jinja
+
+Note: This does *not* override the default value.
+Rather, the value from the pillar/config is prepended to the default.
 
--- a/pillar.example
+++ b/pillar.example
@@ -43,6 +43,8 @@ template:
     # dirs:
     #   files: files_alt
     #   default: default_alt
+    # The entries under `source_files` are prepended to the default source files
+    # given for the state
     # source_files:
     #   template-config-file-file-managed:
     #     - 'example_alt.tmpl'
--- a/template/config/file.sls
+++ b/template/config/file.sls
@@ -13,7 +13,7 @@ include:
 template-config-file-file-managed:
   file.managed:
     - name: {{ template.config }}
-    - source: {{ files_switch(['example.tmpl', 'example.tmpl.jinja'],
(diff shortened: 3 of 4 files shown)


myii:
@baby-gnu If this implementation is accepted, I'll make the small changes necessary to the main TOFS document as well.

myii:
@baby-gnu Actually, I've just gone ahead and done added the changes to the documentation as well.

baby-gnu:
Thanks, I'm reviewing it.

n-rodriguez:
> ```diff
>             - name: /etc/template-formula.conf
>             - source: 
>               - salt://template/files/any/path/can/be/used/here/example.tmpl.jinja
> -             - salt://template/files/ed264f5f1c9d/example.tmpl.jinja
> +             - salt://template/files/any/path/can/be/used/here/example.tmpl
> +             - salt://template/files/any/path/can/be/used/here/example.tmpl.jinja
> +             - salt://template/files/cae0aef658c3/example.tmpl.jinja
> +             - salt://template/files/cae0aef658c3/example.tmpl
> +             - salt://template/files/cae0aef658c3/example.tmpl.jinja
> +             - salt://template/files/Debian-9/example.tmpl.jinja
> +             - salt://template/files/Debian-9/example.tmpl
>               - salt://template/files/Debian-9/example.tmpl.jinja
>               - salt://template/files/Debian/example.tmpl.jinja
> +             - salt://template/files/Debian/example.tmpl
> +             - salt://template/files/Debian/example.tmpl.jinja
>               - salt://template/files/Debian/example.tmpl.jinja
> +             - salt://template/files/Debian/example.tmpl
> +             - salt://template/files/Debian/example.tmpl.jinja
> +             - salt://template/files/default/example.tmpl.jinja
> +             - salt://template/files/default/example.tmpl
>               - salt://template/files/default/example.tmpl.jinja
>             - mode: 644
> ```

@myii :
<Out of subject> do you think this could be done in pillar stack? <Out of subject>
I know macros work in pillar stack since I use them in a private project ;)
Thanks!

myii:
> Out of subject: do you think this could be done in pillar stack?
> I know macros work in pillar stack since I use them in a private project ;)

@n-rodriguez I'm sorry, I don't quite understand the question.  I use PillarStack as well, so I'll be able to check what you're asking.  Do you mean supplying different values to be processed by `files_switch`?  Or actually avoiding TOFS `files_switch` entirely and making a PillarStack-based `files_switch`?

n-rodriguez:
>  Do you mean supplying different values to be processed by `files_switch`? Or actually avoiding TOFS `files_switch` entirely and making a PillarStack-based `files_switch`?

I mean implementing `files_switch` in pillar stack. But it might be confused in my mind too ^^ Need to sleep on it ;)

myii:
@baby-gnu Resolved, to ensure that `example.tmpl.jinja` remains the priority, since that's required for testing.  The (simplified) diff now becomes:

```diff
--- before
+++ after
@@ -1,4 +1,7 @@
     - source:
       - salt://template/files/ABC/example.tmpl.jinja
+      - salt://template/files/ABC/example.tmpl
       - salt://template/files/Debian/example.tmpl.jinja
+      - salt://template/files/Debian/example.tmpl
       - salt://template/files/default/example.tmpl.jinja
+      - salt://template/files/default/example.tmpl
```

myii:
@baby-gnu If you're happy with this, please go ahead and merge it.  I think everyone else leaves TOFS for us two to play with!  Meaning, I don't think anyone else is waiting to review this...

myii:
> I mean implementing `files_switch` in pillar stack. But it might be confused in my mind too ^^ Need to sleep on it ;)

@n-rodriguez I can definitely see the angle on this.  In a way, it would be nice to defer the processing to PillarStack instead.  The biggest problem I can see is that it makes a dependency on (having to configure) PillarStack.  And we're trying to avoid all dependencies for our formulas, at least at the current time.

Right now, I've got three components doing the same sort of thing but at different levels in the system:

1. PillarStack
1. TOFS
1. `stack-formula` (for providing YAML-based SDB configurations, that are accessed using `config.get`)

Combining all of these to use the same processing and directory structure?  That would be very interesting.

myii:
Thanks for the review and merge @baby-gnu.  I'm about to propagate these changes to all of the formulas.  I already started (https://github.com/saltstack-formulas/chrony-formula/pull/23) but now I'll include this as well in one go.

PR #153, opened 2019-07-30
chore: use `semantic-release` cross-formula standard structure

myii:
* Automated using `ssf-formula` (v0.1.0-rc.3)


myii:
Separating this PR from #148 since these changes have already been propagated to all of the other `semantic-release` formulas.  Doesn't make sense to have the `template-formula` "behind" all of these, therefore self-merging this.  #148 will be specific to replacing EOL images.

PR #155, opened 2019-07-31
chore: use `semantic-release` cross-formula standard structure

myii:
* Automated using `ssf-formula` (v0.1.0-rc.4)


myii:
Self-merge since update to documentation after merge of https://github.com/saltstack-formulas/dhcpd-formula/pull/31.

PR #163, opened 2019-09-06
ci: use `dist: bionic` & apply `opensuse-leap-15` SCP error workaround

myii:
* Automated using https://github.com/myii/ssf-formula/pull/22
* Also:
  - https://github.com/myii/ssf-formula/pull/21

---

The [Ubuntu Bionic 18.04](https://docs.travis-ci.com/user/reference/bionic/) build environment is available for Travis CI.  Our current usage patterns are based on [Ubuntu Xenial 16.04](https://docs.travis-ci.com/user/reference/linux).  The plan here is to prefer the latest build environments available and spread this throughout our (`semantic-release`) formulas.

I've actually already run this across the formulas based on the link above.  This has thrown up an ongoing problem that would be really good to solve at the same time: `opensuse-leap-15` regularly fails in almost half of the formulas and always with the same error:

* `SCP did not finish successfully (): (Net::SCP::Error)`

Usually, a single restart gets around this.  Sometimes, even multiple restarts aren't enough.  We really need to solve this or consider dropping back to `opensuse-leap-42` in the meantime.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # vim: ft=yaml
 ---
+dist: bionic
 stages:
   - test
   - lint
@@ -10,7 +11,6 @@ stages:
 sudo: required
 cache: bundler
 language: ruby
-dist: xenial
 
 services:
   - docker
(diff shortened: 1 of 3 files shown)


aboe76:
@myii thanks for this

myii:
@aboe76 Thanks for the merge, I haven't seen an SCP failure yet after this workaround (across all of the formulas), so looking good.

PR #168, opened 2019-09-18
fix(docs): remove old unused documentation files

baby-gnu:
They are only used for Read the Docs, which we're almost definitely
not going to use in the end.


baby-gnu:
See the following comments from @daks and @myii:

* https://github.com/saltstack-formulas/libvirt-formula/pull/43#pullrequestreview-253285365
* https://github.com/saltstack-formulas/libvirt-formula/pull/43#discussion_r297048103
* https://github.com/saltstack-formulas/libvirt-formula/pull/43#discussion_r297051228
* https://github.com/saltstack-formulas/libvirt-formula/pull/43#discussion_r297051455

baby-gnu:
> Can we just leave this for the time being until we can arrange the alternative?

As you want, but it confuse every new user how look at `template-formula` since this formula is a model :-/

myii:
Maybe I'll have to use my fork as the source of the docs.

daks:
> > Can we just leave this for the time being until we can arrange the alternative?
> 
> As you want, but it confuse every new user how look at `template-formula` since this formula is a model :-/

I already tried to delete them, @myii don't want :)

daks:
in fact, having only a link to jump directly to the 'available states' and not the entire list of them seems sufficient for me

Issue #70, opened 2019-03-01
Complete the remaining items in the Community profile


myii:
I.e. https://github.com/saltstack-formulas/template-formula/community.

myii:
Likewise, this has appeared recently: https://github.com/saltstack-formulas/template-formula/security/advisories.

myii:
This has been effectively taken care of org-wide by the [`.github`](https://github.com/saltstack-formulas/.github) repo.

PR #171, opened 2019-09-20
docs(issues): provide `Bug report` & `Feature request` templates

myii:
Notes:

* Related issues:
  - #70 
  - https://github.com/saltstack-formulas/exim-formula/issues/8
* Prepared using:
  - Upstream `salt` issue template
  - Basic template suggested by GitHub

---

The plan is to use across all formulas, if approved.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,43 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: ''
+assignees: ''
+
+---
+
+<!--
+Notes:
+1. Only post _bug reports_ here.
+2. Use the appropriate template for _feature requests_.
+3. Please direct questions to the [`#formulas` channel on Slack](https://saltstackcommunity.slack.com/messages/C7LG8SV54/), which is bridged to `#saltstack-formulas` on Freenode.
+-->
+
+#### Describe the bug
+<!-- A clear and concise description of what the bug is. -->
+
+
+
+#### Setup
+<!-- Provide links to the SLS files and/or relevant configs (be sure to remove sensitive info). -->
+
+
+
+#### Steps to reproduce the bug
+<!-- Include debug logs if possible and relevant. -->
+
+
+
+#### Expected behaviour
+<!-- A clear and concise description of what you expected to happen. -->
+
+
+
+#### Versions Report
+<!-- Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions. -->
+
+
+
+#### Additional context
+<!-- Add any other context about the problem here. -->
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,7 +1,7 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: "[BUG]"
+title: '[BUG] '
 labels: 'bug'
 assignees: ''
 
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+
+#### Setup
+<!-- Provide links to the SLS files and/or relevant configs (be sure to remove sensitive info). -->
+
+
+
+#### Steps to reproduce the bug
+<!-- Include debug logs if possible and relevant. -->
```
daks: Can be interesting to explain how to 'debug log', like running salt with option `-l debug`

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,33 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
```
daks: maybe [FEATURE] or [WISH]

.github/ISSUE_TEMPLATE/bug_report.md
daks: but if we consider user is using Kitchen images, than it's already in debug mode _(i think)_

.github/ISSUE_TEMPLATE/feature_request.md
myii: I actually added that but GitHub lost it (probably because I did both at the same time).

.github/ISSUE_TEMPLATE/bug_report.md
myii: Good idea.

.github/ISSUE_TEMPLATE/feature_request.md
myii: Fixed.

.github/ISSUE_TEMPLATE/bug_report.md
myii: We're currently using `info` rather than `debug`, so I've taken the suggestion and added it.

aboe76:
good idea @myii

myii:
@daks That's an excellent idea but not always the easiest thing to do.  It can also take some time to arrange.  I did that for https://github.com/saltstack-formulas/openvpn-formula/issues/119 and it took me a few attempts to do it.  We could definitely suggest it in the template, though.

myii:
These can also be put under the `docs/` directory instead, to keep things together.  Makes one less directory to manage when transferring across repos.

* https://help.github.com/en/articles/creating-a-pull-request-template-for-your-repository

Any thoughts about that?

aboe76:
@myii if they can be stored in `doc/` please do, that will make the formulas less cluttered.

myii:
@aboe76 Sure, I'm going to do some testing in one of my own repos and make changes here depending on that.  I'll probably add some pull request templates as well, for consideration.

myii:
Unfortunately, I can't get it to work under the `docs` directory.  As soon as I put them under the `.github` directory, they start working again.  GitHub docs aren't particularly useful, since it appears they [changed this feature recently](https://help.github.com/en/articles/manually-creating-a-single-issue-template-for-your-repository).

myii:
Thanks for the reviews, I did make a few minor modifications before the merge.  I'm going to start pushing these out there.  Let me know if there are any issues, I can resolve those relatively easily.

myii:
I've pushed these templates to all of our `semantic-release` formulas.  A few final thoughts here:

1. The PR templates can be worked on next, to complement these templates (i.e. templates for fixing bug reports and adding new features).
1. Once these templates settle down, we can ask for these templates to be managed at an organisational level, so that they are used for all formulas, regardless of `semantic-release` or not.

PR #175, opened 2019-10-08
ci: merge travis matrix, add `salt-lint` & `rubocop` to `lint` job

myii:
* Semi-automated using https://github.com/myii/ssf-formula/pull/60

---

Essentially, this PR is quite simple:

1. Convert the `lint` stage into a job that is run concurrently with the rest of the matrix.
1. Add `salt-lint` to the `lint` job.
1. Add `rubocop` to the `lint` job.

The details are easier to explain commit-by-commit.

---

#### refactor(travis): merge `lint` stage into the `test` stage

* Nothing new here, just a pure refactor to merge the `lint` stage into the `test` stage matrix.
* Even when we first introduced the `lint` stage, there were ideas to run it first, to get the quickest feedback to PR authors and maintainers.
* With the merge into the matrix and making it the first job, we get the best of both worlds.
* https://travis-ci.com/saltstack-formulas/template-formula/builds/130839797.

---

#### ci(travis): run `salt-lint` during the `Lint` job

* This was mentioned by @dafyddj on Slack, starting from:
  - https://freenode.logbot.info/saltstack-formulas/20191004#c2723464
* After a bit of work with the upstream maintainers, it's ready for use with our formulas.
* Basis for the `git ls-files` invocation:
  - Credit: @sblaisot
  - https://freenode.logbot.info/saltstack-formulas/20191004#c2724272

---

#### ci(travis): run `rubocop` during the `Lint` job

* Useful to lint our Ruby files as well (viz. InSpec).

---

#### fix(rubocop): add fixes using `rubocop --safe-auto-correct`

* First set of fixes automated by `rubocop`.

---

#### fix(rubocop): fix remaining errors manually 

* Remaining fixes made manually.


test/integration/default/controls/config_spec.rb
```diff
@@ -6,17 +8,29 @@
...
-    its('content') { should include '"added_in_pillar": "pillar_value"' }
-    its('content') { should include '"added_in_defaults": "defaults_value"' }
-    its('content') { should include '"added_in_lookup": "lookup_value"' }
-    its('content') { should include '"config": "/etc/template-formula.conf"' }
+    its('content') do
+      should include(
+        'This is another example file from SaltStack template-formula.'
+      )
+    end
+    its('content') { should include('"added_in_pillar": "pillar_value"') }
```
dafyddj: Can I point out that the [Inspec style guide](https://www.inspec.io/docs/reference/style/) recommends avoiding unnecessary parentheses in matchers.
> Adding additional parentheses is not required and provides more readability if it is not used
Personally I would agree with that.

.travis.yml
```diff
@@ -1,83 +1,95 @@
...
-    # - INSTANCE: default-opensuse-leap-15-2017-7-py2
-    # - INSTANCE: default-amazonlinux-2-2017-7-py2
-    # - INSTANCE: default-arch-base-latest-2017-7-py2
+# Language and cache config
+language: ruby
+cache: bundler
 
+# Script to run for the test stage
 script:
   - bin/kitchen verify ${INSTANCE}
```
dafyddj:
```suggestion
  - bin/kitchen verify '${INSTANCE}'
```
Adding single quotes here allows for REGEXP targeting in Kitchen, which I have a need for in `vault-formula` (without having to excessively escape the `INSTANCE` var, that is).

test/integration/default/controls/config_spec.rb
myii: @dafyddj Point understood.  My only reason for doing it was for consistency with the lines that had to be split over multiple lines.  Since there's going to be a lot of this throughout the formulas, what method would you suggest?  It's a shame the guide doesn't give guidelines for splitting long lines.

@n-rodriguez Any thoughts, from a general Ruby perspective?

.travis.yml
myii: @dafyddj Ah, that's one of the points I was going to discuss with you in that PR!  While working on standardising our approach across formulas, the thought of using `kitchen list` as the basis for our matrix has crossed my mind many times.  The only thing that has stopped me is the excessive number of entries that can result from it, with `vault-formula` and `iptables-formula` being examples of that.  This wouldn't match up well with the regex method; it's definitely a clever method that needs some consideration (in terms of automation/maintenance).  I'd love to get more opinions about it.

myii: This doesn't take anything away from the fact that quoting `${INSTANCE}` does absolutely no harm at all.  I'll definitely include this and we can continue this discussion in https://github.com/saltstack-formulas/vault-formula/pull/48.

test/integration/default/controls/config_spec.rb
dafyddj: I'm not sure - I'm no Ruby expert. I would need to look into the options.

myii: @dafyddj Since this isn't a critical part of this PR, I'm happy to adjust to the style guide.  I've already made the changes locally and I'll push them here in due course.  Thanks for pointing that out in the first place.

myii: Done.

.travis.yml
myii: Done.

myii: Right, that hasn't worked.

https://travis-ci.com/saltstack-formulas/template-formula/jobs/243467344#L380-L383

```salt
$ bin/kitchen verify '${INSTANCE}'
-----> Starting Kitchen (v2.3.3)
>>>>>> 
No instances for regex `${INSTANCE}', try running `kitchen list'
```

So I've removed that commit and this will have to be covered in https://github.com/saltstack-formulas/vault-formula/pull/48.

dafyddj: Doh! I guess that should be double quotes, but yes, we can leave it for now

.travis.yml
```diff
@@ -1,83 +1,95 @@
...
-    # Define the `lint` stage (runs `yamllint` and `commitlint`)
-    - stage: lint
-      language: node_js
+    # Define the test stage matrix
+    # Run all of the linters in a single instance (concurrently with the main tests)
+    # Define the rest of the matrix based on Kitchen testing
+
+    - language: node_js
       node_js: lts/*
+      env: 'Lint: salt-lint, yamllint, rubocop & commitlint'
```
dafyddj:
```suggestion
      name: 'Lint: salt-lint, yamllint, rubocop & commitlint'
```
What about this (for aesthetic reasons :smiley:)?

myii: @dafyddj I did a lot of digging through Travis docs.  I can't recall the exact place but `env` is used for certain types of control, such as using the `allow_failures` feature.  I've tested that and I'm sure it will come in useful along the way, perhaps where we have certain lint checks that can be allowed to fail.  I haven't checked for `name` specifically, I'm not sure if it will work in the same way.  We can always revisit this later.

dafyddj: Fair enough; `name` just affects how the job appears in the UI, which is what I thought you were going for.

myii: We _could_ use `name` alongside the `env`:

* https://travis-ci.org/myii/template-formula/jobs/595218889

Compared to:

* https://travis-ci.com/saltstack-formulas/template-formula/jobs/243474795

Not really feeling it for the main page, though:

* https://travis-ci.org/myii/template-formula/builds/595218888

I'll unresolve this for the time being, in case there are any comments about this.  Thanks again for the suggestions.

dafyddj: No problem. Either way it's hardly the issue of the century :wink:.

myii: @dafyddj Nope, on reflection, you're right again.  I've updated the PR _and_ I've added a little more info to the `release` stage, so that it's more informative too.  General appearance (slight modifications since then):

* https://travis-ci.org/myii/template-formula/builds/595255179

Great reviewing, by the way.  Appreciate all of the feedback to tidy this up before it gets splurged everywhere!

myii:
So an update: I've run this against every formula using the `ssf-formula`.  They've all tripped up for one reason or another.  Every... single... one!  I was expecting that but to make this process a bit more palatable, I'm going to split each linter into a separate job, so it's easier to find what each formula is failing on.  Then it will be easier to make a plan about how to tackle these lint issues.

---

@n-rodriguez On a separate note, any feedback about the Ruby issue I asked above?

myii:
Yes, that's better.  Most are failing both but some are at least passing `salt-lint`, e.g.:

* https://travis-ci.org/myii/telegraf-formula/builds/595308560
* https://travis-ci.org/myii/locale-formula/builds/595308706
* https://travis-ci.org/myii/keepalived-formula/builds/595308887

Seems like most of the lint fixes are going to be for `rubocop`.

myii:
OK, using `allow_failures` will help get around this problem.  So there will be two types of `.travis.yml` that will be propagated:

1. The ideal: all of the linters in one job, like this PR.
2. Second job: Have a separate job for `rubocop`, which will use `allow_failures`.

I can do that above using `ssf-formula`, probably using the TOFS feature.

We should insist that `salt-lint` passes for every run, so I'll do that with each PR that's created from this process.

@daks @baby-gnu @dafyddj @aboe76 Thanks for taking your time to review this.

n-rodriguez:
@myii 

> @n-rodriguez On a separate note, any feedback about the Ruby issue I asked above?

Sorry for the late answer : usually I conform to https://github.com/rubocop-hq/ruby-style-guide#method-invocation-parens and https://github.com/rubocop-hq/ruby-style-guide#method-invocation-parens-internal-dsl and generally speaking to the whole guide itself.

Between :

```ruby
describe file('/etc/postgresql/9.1/main/pg_hba.conf') do
  its('content') { should match(%r{local\s.*?all\s.*?all\s.*?md5}) }
  its('content') { should match(%r{host\s.*?all\s.*?all\s.*?127.0.0.1\/32\s.*?md5}) }
  its('content') { should match(%r{host\s.*?all\s.*?all\s.*?::1\/128\s.*?md5}) }
end

VS

describe file('/etc/postgresql/9.1/main/pg_hba.conf') do
  its('content') { should match %r{local\s.*?all\s.*?all\s.*?md5} }
  its('content') { should match %r{host\s.*?all\s.*?all\s.*?127.0.0.1\/32\s.*?md5} }
  its('content') { should match %r{host\s.*?all\s.*?all\s.*?::1\/128\s.*?md5} }
end
```
I prefer the second version.

myii:
@n-rodriguez Thanks for the confirmation, that's what I went with in the end.

myii:
@dafyddj @n-rodriguez Interesting note about the parentheses.  In the case with `should match` (i.e. regex), it requires them otherwise it triggers:

* https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/AmbiguousRegexpLiteral

Which probably explains why the InSpec documentation had them in that example.

PR #180, opened 2019-11-22
WIP: Ease reusability of the template

dafyddj:

- [ ] Reviews completed

---

#### Primary type

- [ ] `[fix]`      A bug fix

#### Secondary type

Yes.
All state names and state ids are changed to the new unique token.

### Related issues and/or pull requests

#98 slightly

While creating my first formula using this template, I found it wasn't as straightforward as I'd hoped.
Obviously the word `template` is used all over this formula and some need changing and some don't.

This is just an idea I had of uniquely marking out the replaceable `template` by using the token `template__` - of course, it could be something else too.

Then a simple global replace can be performed while converting the template to a real formula.

Thoughts before I continue further?

### Documentation checklist

- [x] Updated `pillar.example`.

### Testing checklist

- [x] Covered by new/existing tests (e.g. InSpec, Serverspec, etc.).
- [x] Updated the relevant test pillar.


myii:
Thanks for starting this PR, @dafyddj.  It's an interesting take on things, definitely worth a discussion.  I'd like to start by pulling in other contributors who have experience with conversions using the `template-formula` (based on previous discussions).  Hopefully, we can formulate an effective solution to make this process as efficient as possible.

CC: @saltstack-formulas/wg @xenophonf @GMAzrael.

Some alternatives to compare against:

1. Provide a script to convert this formula according to cookiecutter-style input _after_ cloning.
1. Prepare detailed documentation to perform the conversion _after_ cloning.
1. Have a separate repo that's effectively a "pure" template (not numerous things mixed together, as we have in this repo).

myii:
@dafyddj Thanks for the ongoing effort here, this will definitely be useful in resolving this problem.  I hope others can share their opinions, so that we select the best approach going forwards.

myii:
@dafyddj Thanks for the ongoing effort here, this will definitely be useful in resolving this problem.  I hope others can share their opinions, so that we select the best approach going forwards.

daks:
I think that in the middle/long term it should be nice to have a tool/script to generate a formula from scratch like other config management tools have. Maybe cookiecutter can be used for it but I don't really know it so I can't judge if it'll work for our needs.
And maybe we already have a certain stability in this formula to work on it already.

In the meantime _(until someone take this job or we consider this a priority, maybe we already should?)_ manual generation _(like documented in my PR)_ can be nice, and if this PR help it, I'm ok with it.

myii:
@dafyddj The script looks great, this is really taking shape.  Some comments:

* Since our formulas are consumed across various OSes, can we make this as portable as possible (i.e. POSIX standard)?
  - `sh` rather than `bash`
  - [Any alternative](https://unix.stackexchange.com/questions/123144/which-is-the-most-portable-of-sed-awk-perl-and-sh) to `perl`, such as `sed`
* Apply the improvements suggested by https://www.shellcheck.net/
  - Of course, [ignoring](https://github.com/koalaman/shellcheck/wiki/Ignore) wherever necessary
  - As a side note, this is [integrated by default into Travis CI](https://github.com/koalaman/shellcheck#travis-ci), which I think would be a great addition to the lint stage -- thoughts?

myii:
@dafyddj I've started #181, with the plan to propagate using `shellcheck` to all of the formulas.  Ignores are simple enough where required.

bin/convert-formula.sh
```diff
@@ -0,0 +1,46 @@
...
+  fi
+}
+
+convert_formula() {
+  git reset \
+    "$(echo "feat: initial commit" \
+      | git commit-tree "HEAD^{tree}")"
+  git rm bin/convert-formula.sh AUTHORS.md CHANGELOG.md
+  git mv template__  "${NEW_NAME}"
+  grep -rl template__ . --exclude-dir=.git | xargs perl -pi -e s/template__/"${NEW_NAME}"/g
```
myii: @dafyddj This may help with the `perl` vs. `sed` issue:

* https://github.com/saltstack/salt-bootstrap/commit/8647b74cbc7f134b22d8b5ab6f9975440736dc37

dafyddj: Thanks, but the problematic flag is `-i` as explained [here](https://stackoverflow.com/q/5694228/5009408) but as I look closer there may be an another way further down in the answers on that page.

myii:
@dafyddj Thanks for moving over to `sh`.  A quick question, not prescriptive: what are your thoughts on using uppercase `TEMPLATE` instead of `template__`?  I haven't thought this through but it just occurred to me, so I thought I'd ask.

bin/convert-formula.sh
dafyddj: Resolved with `ex`.

dafyddj:
I'm not particularly attached to `template__` and I think `TEMPLATE` could work because the search/replace is case-sensitive. Would we ever have a legitimate need to use TEMPLATE in the formula? Probably not.

Also, do you agree that this conversion process should be CI tested? I'm thinking, in Travis, clone the repo, run the script, then run at least one of the test suites in that cloned repo. Thoughts?

myii:
> Also, do you agree that this conversion process should be CI tested? I'm thinking, in Travis, clone the repo, run the script, then run at least one of the test suites in that cloned repo. Thoughts?

@dafyddj Definitely useful and worthwhile.  A single job would be enough to test that.

dafyddj:
@myii this is mostly done as far as I'm concerned. 

It could probably do with some documentation. Can you suggest where best to place the docs - and should the script remove that part of the documentation when it's done?

myii:
> @myii this is mostly done as far as I'm concerned.

Great, I'm going to take this for a test drive soon, since I need to make a formula or two.  That will help me finalise the review.

> It could probably do with some documentation. Can you suggest where best to place the docs - and should the script remove that part of the documentation when it's done?

The `README` is a good place.  A couple of places to consider:

* A sub-heading under https://github.com/saltstack-formulas/template-formula#special-notes
* Or since this repo is a template, we could use a section right at the very top, before the general notes 

Since the end-user is going to be editing the `README` for their own purposes, I don't think the script needs to remove this.  I'll confirm this after my tests.

bin/convert-formula.sh
```diff
@@ -0,0 +1,52 @@
...
+      | git commit-tree 'HEAD^{tree}')"
+  git rm --quiet bin/convert-formula.sh AUTHORS.md CHANGELOG.md
+  git mv TEMPLATE "${NEW_NAME}"
+  grep --recursive --files-with-matches --exclude-dir=.git TEMPLATE . \
+    | xargs -L 1 ex -sc '%s/TEMPLATE/'"${NEW_NAME}"'/g|x'
+  # shellcheck disable=SC2016 # Expressions don't expand in single quotes
+  git commit --quiet --all \
+    --message 'feat: convert `template-formula` to `'"${NEW_NAME}"'-formula`
+
+BREAKING CHANGE: changed all state names and ids'
```
myii: Suggestion:

```sh
  git commit --quiet --all \
    --message 'feat: convert `template-formula` to `'"${NEW_NAME}"'-formula`' \
    --message 'BREAKING CHANGE: changed all state names and ids'
```

bin/convert-formula.sh
```diff
@@ -0,0 +1,52 @@
...
+  # Empty history and make commit message `semantic-release`-compliant
+  # Works for forks of `template-formula` as well as GitHub "Use this template"
+  # See https://stackoverflow.com/a/15572071/5009408
+  git reset \
+    "$(echo 'feat: initial commit' \
+      | git commit-tree 'HEAD^{tree}')"
+  git rm --quiet bin/convert-formula.sh AUTHORS.md CHANGELOG.md
+  git mv TEMPLATE "${NEW_NAME}"
+  grep --recursive --files-with-matches --exclude-dir=.git TEMPLATE . \
+    | xargs -L 1 ex -sc '%s/TEMPLATE/'"${NEW_NAME}"'/g|x'
```
myii: In my first local tests, the script doesn't get past this line.  Since this is working in Travis, there's probably something in the way I've patched things together.  The only way past it is to comment out line 3, i.e.:

```sh
# set -o errexit # If a command fails exit the whole script
```

dafyddj: You might get some clues by running as (if you haven't already done so):
```DEBUG=true bin/convert-formula.sh ...```

Comment (myii) [reply bin/convert-formula.sh]: Yes, I started with that and then tried all sorts, including removing all of the `quiet` or `silent` flags but to no avail.  Even when I get it running with this line commented out, there are no warning given.  Not sure what's happening with that `ex` command to cause an error status...

Comment (myii) [reply bin/convert-formula.sh]: @dafyddj This sounds a little strange but I found a solution (using `-u NONE`) based on this reference:

* https://unix.stackexchange.com/questions/14497/why-would-vim-return-non-zero-exit-code-if-i-exit-immediately-after-opening

So that's:

```sh
  grep --recursive --files-with-matches --exclude-dir=.git TEMPLATE . \
    | xargs -L 1 ex -u NONE -sc '%s/TEMPLATE/'"${NEW_NAME}"'/g|x'
```

* This actually worked, with `set -o errexit` re-enabled.
* Maybe the end user (like me) could have something strange in their configuration, affecting `ex` when it exits?

bin/convert-formula.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+convert_formula() {
+  # Empty history and make commit message `semantic-release`-compliant
+  # Works for forks of `template-formula` as well as GitHub "Use this template"
+  # See https://stackoverflow.com/a/15572071/5009408
+  git reset \
+    "$(echo 'feat: initial commit' \
+      | git commit-tree 'HEAD^{tree}')"
+  git rm --quiet bin/convert-formula.sh AUTHORS.md CHANGELOG.md
```
myii:
```suggestion
  git rm --quiet bin/convert-formula.sh AUTHORS.md CHANGELOG.md \
    docs/_static/css/custom.css docs/AUTHORS.rst docs/CHANGELOG.rst \
    docs/conf.py docs/CONTRIBUTING_DOCS.rst docs/index.rst
```

myii:
Current status:

1. Discuss and finalise changes as suggested above.
1. Add notes to the `README`, including explaining that `git push -f` needs to be used and perhaps how to configure Travis to enable `semantic-release` (e.g. `GH_TOKEN`, etc.).

A wish-list collection for the long-run, doesn't prevent merging this PR:

1. `README`: Update the headings symbols to match the length of the heading itself.
1. `.rubocop.yml`: Remove the `Metrics/BlockLength` section, specific to the `template-formula`.
1. `.travis.yml`: Remove the `env: 'Conversion'` block.
1. `FORMULA`: reset the `version` to `1.0.0`.

myii:
@dafyddj Merged, the rest of the issues can be covered in a future PR.  Thanks for this fantastic PR.

PR #181, opened 2019-11-27
[ssf] ci(travis): run `shellcheck` during lint job

myii:
* Automated using https://github.com/myii/ssf-formula/pull/106

---

To be propagated across all formulas using the `ssf-formula`.  According to our discussions, this only requires _one approval_, since it's only CI-based changes.

Idea started from here:

* https://github.com/saltstack-formulas/template-formula/pull/180#issuecomment-558612422

As `shellcheck` is already integrated into Travis, this is easy to run and asides from here, already tested elsewhere:

* https://travis-ci.com/saltstack-formulas/fail2ban-formula/jobs/260717687#L215-L216
* https://travis-ci.org/myii/ssf-formula/builds/617730995#L206-L207

An example of failures (`shellcheck` has excellent error descriptions):

* https://travis-ci.org/myii/ssf-formula/builds/617359726#L206-L367

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,13 +32,13 @@ jobs:
     - language: 'node_js'
       node_js: 'lts/*'
       env: 'Lint'
-      name: 'Lint: salt-lint, yamllint, rubocop & commitlint'
+      name: 'Lint: salt-lint, yamllint, rubocop, shellcheck & commitlint'
       before_install: 'skip'
       script:
         # Install and run `salt-lint`
         - pip install --user salt-lint
-        - git ls-files | grep '\.sls$\|\.jinja$\|\.j2$\|\.tmpl$\|\.tst$'
-                       | xargs salt-lint
+        - git ls-files -- *.sls *.jinja *.j2 *.tmpl *.tst
+                        | xargs salt-lint
         # Install and run `yamllint`
         # Need at least `v1.17.0` for the `yaml-files` setting
         - pip install --user yamllint>=1.17.0
@@ -46,6 +46,10 @@ jobs:
         # Install and run `rubocop`
         - gem install rubocop
         - rubocop -d
+        # Run `shellcheck` (already pre-installed in Travis)
+        - shellcheck --version
+        - git ls-files -- *.sh *.bash *.ksh
+                        | xargs shellcheck
         # Install and run `commitlint`
         - npm i -D @commitlint/config-conventional
                    @commitlint/travis-cli
--- a/pre-commit_semantic-release.sh
+++ b/pre-commit_semantic-release.sh
@@ -14,9 +14,9 @@ sed -i -e "s_^\(version:\).*_\1 ${1}_" FORMULA
 sudo -H pip install m2r
 
 # Copy and then convert the `.md` docs
-cp *.md docs/
-cd docs/
-m2r --overwrite *.md
+cp ./*.md docs/
+cd docs/ || exit
+m2r --overwrite ./*.md
 
 # Change excess `H1` headings to `H2` in converted `CHANGELOG.rst`
 sed -i -e '/^=.*$/s/=/-/g' CHANGELOG.rst


.travis.yml
```diff
@@ -46,6 +46,10 @@ jobs:
         # Install and run `rubocop`
         - gem install rubocop
         - rubocop -d
+        # Run `shellcheck` (already pre-installed in Travis)
+        - shellcheck --version
+        - git ls-files | grep '\.sh$\|\.bash$\|\.ksh$'
```
dafyddj: Do you know that `git ls-files` accepts patterns itself as [pathspecs](https://kgrz.io/git-intro-to-pathspec.html), so
```
git ls-files -- '*.sh' '*.bash' '*.ksh' | xargs shellcheck
```
would work equally well here (and reads better imho)?

myii: @dafyddj Good to know, looks like it works without the quotes as well.  Will have to update the `salt-lint` invocation as well when applying this.

myii: Applied, thanks.

myii:
Thanks for the reviews/comments, starting the propagation of this now.

.travis.yml
dafyddj: > @dafyddj Good to know, looks like it works without the quotes as well. Will have to update the `salt-lint` invocation as well when applying this.

To my knowledge it shouldn't work _correctly_ without quotes. How did you test that?

myii: @dafyddj The simplest check, comparing the output of both without the pipe to `shellcheck` (or `salt-lint` in the example below), i.e.:

```bash
fail2ban-formula$  git ls-files -- '*.sls' '*.jinja' '*.j2' '*.tmpl' '*.tst'
fail2ban/config.sls
fail2ban/deprecated.sls
fail2ban/init.sls
fail2ban/map.jinja
fail2ban/ng/config.sls
fail2ban/ng/files/config.jinja
fail2ban/ng/init.sls
fail2ban/ng/install.sls
fail2ban/ng/map.jinja
fail2ban/ng/service.sls
test/salt/states/misc/fake_log_files.sls
fail2ban-formula$  git ls-files -- *.sls *.jinja *.j2 *.tmpl *.tst
fail2ban/config.sls
fail2ban/deprecated.sls
fail2ban/init.sls
fail2ban/map.jinja
fail2ban/ng/config.sls
fail2ban/ng/files/config.jinja
fail2ban/ng/init.sls
fail2ban/ng/install.sls
fail2ban/ng/map.jinja
fail2ban/ng/service.sls
test/salt/states/misc/fake_log_files.sls
```

Also tested this out with the pipes to both `shellcheck` and `salt-lint` respectively.

myii: @dafyddj Never mind, I've already started propagating the quotes using:

* https://github.com/myii/ssf-formula/pull/107

It should be done within the next few minutes.  Thanks for getting back to me.

dafyddj: I don't deny that your test works but there's something strange going on here for me, and I'm trying to work out what that is.

dafyddj: Ok, your test only worked by accident. :)

From [gnu](https://www.gnu.org/software/bash/manual/html_node/Filename-Expansion.html)

> If no matching filenames are found, and the shell option nullglob is disabled, the word is left unchanged.

So because no files happen to match in the root directory of `fail2ban-formula` `git ls-files` receives the patterns themselves, and works as expected. The results would not be complete if a matching file existed in the root directory.

To be honest, I've never come across this situation before so it was handy to learn more about Bash!

myii: Funny, because it worked in the first few formulas I tried!  At least you got to the bottom of that, thanks for sharing.

Actually, this overlaps with the example here, using `shopt` to keep `nullglob` enabled:

* https://github.com/koalaman/shellcheck/wiki/TravisCI

dafyddj:
Interesting. That explains a lot.

PR #195, opened 2020-04-25
chore(defaults): add default retry dict (WIP)

noelmcloughlin:

- [ ] Reviews completed

---

#### Primary type

- [x] `[chore]`    Changes to the build process or auxiliary tools and libraries such as documentation generation
- [ ] `[fix]`      A bug fix

#### Secondary type

No.

Let's add a retry Dict for fast implementation of retries.  SLS usage is:   

                         `retry: {{ TEMPLATE.retry_option }}`

### Documentation checklist

- [x] Updated the `README` (e.g. `Available states`).

### Testing checklist

### DIFF
--- a/TEMPLATE/defaults.yaml
+++ b/TEMPLATE/defaults.yaml
@@ -13,3 +13,10 @@ TEMPLATE:
   # Just here for testing
   added_in_defaults: defaults_value
   winner: defaults
+
+  retry_option:
+    # https://docs.saltstack.com/en/latest/ref/states/requisites.html#retrying-states
+    attempts: 3
+    until: true
+    interval: 10
+    splay: 10


noelmcloughlin:
It's appears as `retry_options` in [some formulas](https://github.com/search?q=org%3Asaltstack-formulas+retry_options&type=Code) but I always use `retry_option` because that's what the feature is called in [salt documentation](https://docs.saltstack.com/en/latest/ref/states/requisites.html#retrying-states). 

> The retry option in a state

From a state perspective, there is only one **retry option** not multiple **retry options.**

So the main question is whether `s` is preferred in this PR.

noelmcloughlin:
And [some formulas](https://github.com/search?q=org%3Asaltstack-formulas+retry_option&type=Code) use `retry_option`

daks:
If I understand it correctly this PR had a the option to run each state 3 times until it succeeds.

Not sure I'm in favour, a state should succeed at the first apply, if it does not I prefer to know it.
_(during some time, our kitchen setup also ran highstate twice because first one never succeeded. We didn't see the problems until we changed this setup)_

noelmcloughlin:
This PR adds an example default implementation of salt requisites, retry:
https://docs.saltstack.com/en/latest/ref/states/requisites.html#retrying-states
useful when a state author wants to add retry option to that state. Retries
are really useful for *archive.extracted* when source=internet, and during
development using hotel wifi ;-). I would use a retry option selectively.
Both retry>1 and retry=0 are both valid in reactive systems approach where
resilience is important, but this PR does not use or force the feature.

daks:
So, this adds the default option but it still needs to be *explicitely* added to each state?

noelmcloughlin:
Yes. But we could put the PR on hold. Maybe the template-formula should
include an example usage?

noelmcloughlin:
Closing as stale.

PR #196, opened 2020-04-25
feat(repo): basic linux repo support (WIP)

noelmcloughlin:

- [ ] Reviews completed

---

#### Primary type

- [ ] `[fix]`      A bug fix

#### Secondary type

No.

Formula users seem to often have a desire or requirement to install software from an upstream repo instead of using default OS package.  Searching for the term `pkgrepo.managed` returns multiple hits on the saltstack-formulas repo:  https://github.com/search?q=org%3Asaltstack-formulas+pkgrepo.managed&type=Code 

Since this is such a common pattern we should showcase this in template-formula gradually. 

This PR is basic support for a `pkg.repo` dictionary and install/clean states which are linked to a `pkg.use_upstream_repo` flag.

If some other implementation is preferred let me know - I know there are various discussions about stuff in the core team which maybe I'm not up to speed on.

### Documentation checklist

### Testing checklist

### DIFF
--- a/TEMPLATE/package/install.sls
+++ b/TEMPLATE/package/install.sls
@@ -1,20 +1,18 @@
 # -*- coding: utf-8 -*-
 # vim: ft=sls
 
-{#- Get the `tplroot` from `tpldir` #}
 {%- set tplroot = tpldir.split('/')[0] %}
 {%- set sls_repo = tplroot ~ '.repo.install' %}
 {%- from tplroot ~ "/map.jinja" import TEMPLATE with context %}
 
-         {%- if TEMPLATE.pkg.use_upstream_repo %}
+{%- set requires = [] %}
+{%- if TEMPLATE.pkg.use_upstream_repo %}
+       {%-   set requires.append({'sls': sls_repo}) %}
 include:
   - {{ sls_repo }}
-         {%- endif %}
+{%- endif %}
 
 TEMPLATE-package-install-pkg-installed:
   pkg.installed:
     - name: {{ TEMPLATE.pkg.name }}
-         {%- if TEMPLATE.pkg.use_upstream_repo %}
-    - require:
-      - sls: {{ sls_repo }}
-         {%- endif %}
+    - require: {{ requires }}
--- a/TEMPLATE/package/install.sls
+++ b/TEMPLATE/package/install.sls
@@ -3,8 +3,18 @@
 
 {#- Get the `tplroot` from `tpldir` #}
 {%- set tplroot = tpldir.split('/')[0] %}
+{%- set sls_repo = tplroot ~ '.repo.install' %}
 {%- from tplroot ~ "/map.jinja" import TEMPLATE with context %}
+
+         {%- if TEMPLATE.pkg.use_upstream_repo %}
+include:
+  - {{ sls_repo }}
+         {%- endif %}
 
 TEMPLATE-package-install-pkg-installed:
   pkg.installed:
     - name: {{ TEMPLATE.pkg.name }}
+         {%- if TEMPLATE.pkg.use_upstream_repo %}
+    - require:
+      - sls: {{ sls_repo }}
+         {%- endif %}
--- a/TEMPLATE/defaults.yaml
+++ b/TEMPLATE/defaults.yaml
@@ -2,8 +2,6 @@
 # vim: ft=yaml
 ---
 TEMPLATE:
-  pkg:
-    name: TEMPLATE
   rootgroup: root
   config: '/etc/TEMPLATE'
   service:
@@ -13,3 +11,15 @@ TEMPLATE:
   # Just here for testing
   added_in_defaults: defaults_value
   winner: defaults
+
+  pkg:
+    name: TEMPLATE
+    use_upstream_package: true  # reserve for future use
+    use_upstream_repo: false    # managed by os(family)maps
+    repo:                       # managed by os(family)maps
+      humanname: TEMPLATE
+      name: TEMPLATE
+      comments:
+        - installed by salt
+      enabled: 1
+      gpgcheck: 1
--- a/TEMPLATE/files/macros.jinja
+++ b/TEMPLATE/files/macros.jinja
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# vim: ft=jinja
+#
+# Collection of common macros
+
+{%- macro format_kwargs(kwarg) -%}
+
+  {%- filter indent(4) %}
+    {%- for k, v in kwarg|dictsort() %}
+- {{ k }}: {{ v }}
+    {%- endfor %}
+  {%- endfilter %}
+
+{%- endmacro %}
--- a/TEMPLATE/osfamilymap.yaml
+++ b/TEMPLATE/osfamilymap.yaml
@@ -11,9 +11,14 @@
 # osfamilymap: {}
 ---
 Debian:
+  config: /etc/TEMPLATE.d/custom.conf
   pkg:
     name: TEMPLATE-debian
-  config: /etc/TEMPLATE.d/custom.conf
+    use_upstream_repo: true
+    repo:
+      name: deb https://apt.TEMPLATE.io/ TEMPLATE-xenial main
+      file: /etc/apt/sources.list.d/TEMPLATE.list
+      key_url: https://packages.cloud.google.com/apt/doc/apt-key.gpg
 
 RedHat:
   pkg:
--- a/TEMPLATE/package/clean.sls
+++ b/TEMPLATE/package/clean.sls
@@ -1,19 +1,21 @@
 # -*- coding: utf-8 -*-
 # vim: ft=sls
 
-{#- Get the `tplroot` from `tpldir` #}
 {%- set tplroot = tpldir.split('/')[0] %}
 {%- set sls_config_clean = tplroot ~ '.config.clean' %}
 {%- set sls_repo_clean = tplroot ~ '.repo.clean' %}
 {%- from tplroot ~ "/map.jinja" import TEMPLATE with context %}
 
+{%- set requires = [] %}
+{%- set requires.append({'sls': sls_config_clean}) %}
 include:
   - {{ sls_config_clean }}
+{%- if TEMPLATE.pkg.use_upstream_repo %}
+       {%-   set requires.append({'sls': sls_repo_clean}) %}
   - {{ sls_repo_clean }}
+{%- endif %}
 
 TEMPLATE-package-clean-pkg-removed:
   pkg.removed:
     - name: {{ TEMPLATE.pkg.name }}
-    - require:
-      - sls: {{ sls_config_clean }}
-      - sls: {{ sls_repo_clean }}
+    - require: {{ requires }}
--- a/TEMPLATE/package/clean.sls
+++ b/TEMPLATE/package/clean.sls
@@ -4,13 +4,16 @@
 {#- Get the `tplroot` from `tpldir` #}
 {%- set tplroot = tpldir.split('/')[0] %}
 {%- set sls_config_clean = tplroot ~ '.config.clean' %}
+{%- set sls_repo_clean = tplroot ~ '.repo.clean' %}
 {%- from tplroot ~ "/map.jinja" import TEMPLATE with context %}
 
 include:
   - {{ sls_config_clean }}
+  - {{ sls_repo_clean }}
 
 TEMPLATE-package-clean-pkg-removed:
   pkg.removed:
     - name: {{ TEMPLATE.pkg.name }}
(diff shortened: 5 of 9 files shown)


TEMPLATE/package/install.sls
```diff
@@ -3,8 +3,18 @@
...
+  - {{ sls_repo }}
+         {%- endif %}
+
 TEMPLATE-package-install-pkg-installed:
   pkg.installed:
     - name: {{ TEMPLATE.pkg.name }}
+         {%- if TEMPLATE.pkg.use_upstream_repo %}
+    - require:
+      - sls: {{ sls_repo }}
+         {%- endif %}
```
baby-gnu, line 20: Why not building a `require` jinja variable at include time?

Maybe like:

```sls
{%-   set requires = [] %}
{%- if TEMPLATE.pkg.use_upstream_repo %}
{%-   set requires.append({'sls': sls_repo}) %}
include:
  - {{ sls_repo }}
{%- endif %}

TEMPLATE-package-install-pkg-installed:
  pkg.installed:
    - name: {{ TEMPLATE.pkg.name }}
    - require: {{ requires }}
```

I think we shoud try to avoid conditionnal as much as possible in the middle of a state as [explained in conventions](https://docs.saltstack.com/en/latest/topics/development/conventions/formulas.html#light-conditionals-and-looping).

noelmcloughlin: I like this suggestion @baby-gnu It's both pythonic and jinjaic ;-)
Yes, I don't like including too much Jinja logic in SLS states.

noelmcloughlin: However, I'm not sure [conditionals-and-looping ](https://docs.saltstack.com/en/latest/topics/development/conventions/formulas.html#light-conditionals-and-looping ) applies to this use case.

The install.sls looks okay, just one include.
```
{%- set requires = [] %}
{%- if TEMPLATE.pkg.use_upstream_repo %}
       {%-   set requires.append({'sls': sls_repo}) %}
include:
  - {{ sls_repo }}
{%- endif %}
```
But the clean.sls becomes messy - multiple includes
```
{%- set requires = [] %}
{%- set requires.append({'sls': sls_config_clean}) %}
include:
  - {{ sls_config_clean }}
{%- if TEMPLATE.pkg.use_upstream_repo %}
       {%-   set requires.append({'sls': sls_repo_clean}) %}
  - {{ sls_repo_clean }}
{%- endif %}
```

noelmcloughlin: Some kind of syntax issue?
```
           Rendering SLS 'base:TEMPLATE.package.install' failed: Jinja syntax error: expected token 'end of statement block', got '('; line 10

       {%-   set requires.append({'sls': sls_repo}) %}    <======================
```

baby-gnu: I did not test it ;-)

ho yes, the problem is `set` it should be a `do`:

```
{%-    do requires.append({'sls': sls_repo}) %}
```

I think the same idea could be done for the include in `clean.sls`:

``` sls
{%- set includes = [sls_config_clean] %}
{%- set requires = [{'sls': sls_config_clean}] %}
{%- if TEMPLATE.pkg.use_upstream_repo %}
       {%-   do includes.append(sls_repo_clean) %}
       {%-   do requires.append({'sls': sls_repo_clean}) %}
{%- endif %}

include: {{ includes }}
```

It could be possible to deduce `include` from `requires` using jinja [map](https://jinja.palletsprojects.com/en/2.11.x/templates/#map), the following is untested

``` sls
{%- set requires = [{'sls': sls_config_clean}] %}
{%- if TEMPLATE.pkg.use_upstream_repo %}
       {%-   do requires.append({'sls': sls_repo_clean}) %}
{%- endif %}

include: {{ requires | map(attribute='sls') }}
```

Regards.

daks:
Not sure I understand why you want to introduce this in the template-formula. 

Do you consider it an absolute pattern which must be present in all derivative _(meaning: as created from this one)_ formulas?

javierbertoli:
Sorry to jump in, but while working on another formula that uploaded yesterday, I added some [repo management code](https://github.com/saltstack-formulas/arvados-formula/tree/master/arvados/repo) that was considering to submit to the template. It does things a little different to this one. 

Not sure if something worth can be picked from that one and merged in this PR?

noelmcloughlin:
> Sorry to jump in, but while working on another formula that uploaded yesterday, I added some [repo management code](https://github.com/saltstack-formulas/arvados-formula/tree/master/arvados/repo) that was considering to submit to the template. It does things a little different to this one.
> 
> Not sure if something worth can be picked from that one and merged in this PR?

I think your code is similar to my original PR.  Instead of having IF Debian/RedHat code in the SLS, the OS-specific parameters should be in osfamilymap and pulled in using macro IMHO.  I have generally followed the **Postgres-formula** pattern of repo handling.  Your avardos Jinja is cleaner, but I was just following the existing template-formula usage of requires SLS.

It just occurs to me that `requires: SLS: install-some-software` is inflexible. For example, I use  [ceph-formula .](https://github.com/saltstack-formulas/ceph-formula/blob/master/ceph/repo/init.sls) to setup the CEPH repository **only**   ... I don't want to use that formula to install ceph.   Setting up a repo does not require other states - its standalone task (in theory).

noelmcloughlin:
> Do you consider it an absolute pattern which must be present in all derivative _(meaning: as created from this one)_ formulas?

I'm not sure. The current template-formula is heavily biased towards RPM/YUM/etc but most technology software is distributed as archives/tarballs in the first instance.  

At the moment when I want a template for repo handling, I copy some other formula because template-formula has no template for repo handling. I usually add repo handling because end users want this - it's not a feature I use myself, I prefer tarballs.  Maybe I'll use snaps in future.

daks:
> > Do you consider it an absolute pattern which must be present in all derivative _(meaning: as created from this one)_ formulas?
> 
> I'm not sure. The current template-formula is heavily biased towards RPM/YUM/etc but most technology software is distributed as archives/tarballs in the first instance.
> 
> At the moment when I want a template for repo handling, I copy some other formula because template-formula has no template for repo handling. I usually add repo handling because end users want this - it's not a feature I use myself, I prefer tarballs. Maybe I'll use snaps in future.

I understand the need to the 'pattern' on some (or maybe even most) formulas. I'm just wondering if we want to 'preload' `template-formula` with all utilities/patterns/useful code which may be needed to derivative formulas or not.
It may have sense if we want template-formula to be the unique source and as a reference of 'salt best practices'. I'm just concerned *(as often :)* about the result in complexity of its code and so in resulting difficulty for new users to understand and choose what to keep and what to discard when creating a formula from this one.

myii:
@daks I know @dafyddj isn't too keen on having an interactive script but how about if we could add some interaction to get rid off sections selectively?  Of course, we could use `--non-interactive` by default and only provide these extra options if someone passes the `--interactive` switch.

javierbertoli:
> @daks I know @dafyddj isn't too keen on having an interactive script but how about if we could add 
> some interaction to get rid off sections selectively? Of course, we could use --non-interactive by 
> default and only provide these extra options if someone passes the --interactive switch.

I was thinking on the same lines, not interactive, but with a switch for the  `convert-formula.sh` script, like:

```bash
convert-formula.sh                               # current behaviour
                                                 # (package, config, service)
convert-formula.sh --package-only                # trims the formula for package
                                                 # management only + repo management
convert-formula.sh --package-only --without-repo # trims the formula for package
                                                 # management only and removes
                                                 # the repo management code
```
That way, the user can choose what to keep or not?

noelmcloughlin:
Yes. I agree overloading with patterns is a dilemma which I have seen
discussed elsewhere whereby maybe this formula should have multiple
branches.  Do you know if that discussion came to a conclusion?

If we load all this stuff on the master branch that may be okay.  In
reality, the main stakeholders for template formula are the core team
because we are most likely to add new formulas.  But we could maintain our
own branches or forks for that.

 I could not clone the template-formula when creating some formulas
recently because it excluded useful stuff like retries, archive handling,
repos, and more.  But maybe we hold the PR until we decide what we want
template formula structure to be, and figure out who is primary audience
for template-formula - I suspect the primary audience is us , the main
contributors, even though we try to convince ourselves the audience is new
users ;-).

To re-answer your original question, No, this is not absolute necessary
pattern.

javierbertoli:
When I helped maintain a [consistent set of puppet modules](https://github.com/example42/Example42-templates) that I found around and liked (pretty much what happened with these formulas :smile: ), they had a [few different templates that were maintained](https://github.com/example42/Example42-templates#modules-templates) for each case.

It seems to me that there's no '_one-solution-fits-all_' answer to this and we'll need to choose our poison:

* a **beefy template**, with all the bells and whistles, and then let the consumer trim the clone (either with a script or manually)
* a **bunch of templates**, for each case, as in the puppet link above, harder to maintain, easier to consume. I think that, in this case, we can create the beefy template as above, and  **we** can run the script, as proposed, to maintain other 'simpler' templates? 
* a **minimal template**, with snippets/examples to be added (I guess this is the least reasonable solution)

Agreed, it's quite possible we'll be the main consumers of the templates, but nonetheless, we should consider other people exist out there that might want to use them, too. Right? :smile:

noelmcloughlin:
I think formulas could include `examples/` directory with samples. With Docker and Vagrant, I can upload docker and vagrant files representing example configuration I am promoting. I cannot do this with saltstack formulas because the only file available for this is `pillar.example` which is sometimes unusable without modification.

Basically the `examples/` directory could contain files named and/or contributed by `noel`, `javier`, `daks`, `dafyddj`, `baby-gnu`, `myii` for example. The end-user could picks one that fits their use case, or contribute their own.  We are end users too and sometimes I have preferred configuration which I currently store somewhere else, despite being directly related to the formula.

myii:
> I think formulas could include `examples/` directory with samples. With Docker and Vagrant, I can upload docker and vagrant files representing example configuration I am promoting. I cannot do this with saltstack formulas because the only file available for this is `pillar.example` which is sometimes unusable without modification.
> 
> Basically the `examples/` directory could contain files named and/or contributed by `noel`, `javier`, `daks`, `dafyddj`, `baby-gnu`, `myii` for example. The end-user could picks one that fits their use case, or contribute their own. We are end users too and sometimes I have preferred configuration which I currently store somewhere else, despite being directly related to the formula.

This is actually a very interesting idea and goes along nicely with `Usage` section we've been discussing in the Formulas' working group meetings (cc: @baby-gnu):

* https://tree.taiga.io/project/myii-saltstack-formulas/task/86

Thanks for sharing this, @noelmcloughlin.

noelmcloughlin:
Closing - the idea can be picked up again if needed.

TEMPLATE/package/install.sls
baby-gnu: > It could be possible to deduce `include` from `requires` using jinja [map](https://jinja.palletsprojects.com/en/2.11.x/templates/#map), the following is untested

After some tests I found 2 issues:
1. `map` return an iterator which could not be used directly in `include`, we need the `list` filter
2. it may have some encoding error with Python2

Here is a working example:

```sls
{#- Get the `tplroot` from `tpldir` #}
{%- set tplroot = tpldir.split("/")[0] %}
{%- set sls_repo = tplroot ~ ".repo.install" %}
{%- from tplroot ~ "/map.jinja" import TEMPLATE with context %}


{%- if TEMPLATE.pkg.use_upstream_repo %}
{%-   set requires = [{"sls": sls_repo}] %}
{%- else %}
{%-   set requires = [] %}
{%- endif %}

{#- serialize as YAML to avoid encoding errors on Python2 #}
include: {{ requires | map(attribute="sls") | list | yaml }}

TEMPLATE-package-install-pkg-installed:
  pkg.installed:
    - name: {{ TEMPLATE.pkg.name }}
    - require: {{ requires | yaml }}
```

myii: @baby-gnu I reckon our policy should be to completely ignore Python 2 nowadays.  We've already got enough on our plate and it would be great to streamline formulas as much as possible.

baby-gnu: My problem is that the [salt-bootstrap](https://github.com/saltstack/salt-bootstrap) script installed the python2 version by default on a Bionic VM :-/

noelmcloughlin: That should be fixed. Meanwhile, you can pass `-x python3` to the `salt-bootstrap` command line to get that fixed.

baby-gnu: > That should be fixed. Meanwhile, you can pass `-x python3` to the `salt-bootstrap` command line to get that fixed.

Yes, having some troubles about Jinja which could not find a `mapdata.foo` because the `foo` was `mapdata[u'foo']` I finally thought about the python2 issue and found the `-x python3` option.

PR #203, opened 2020-08-31
test(map): verify `map.jinja` dump using `_mapdata` state

baby-gnu:

- [x] Reviews completed

---

#### Primary type

- [x] `[ci]`       Changes to the continuous integration configuration
- [ ] `[fix]`      A bug fix

#### Secondary type

- [x] `[test]`     Adding missing or correcting existing tests

No.

Verify `map.jinja` dump using `_mapdata` state.

### Documentation checklist

### Testing checklist

- [x] Covered by new/existing tests (e.g. InSpec, Serverspec, etc.).

### DIFF
--- a/TEMPLATE/_mapdata/init.sls
+++ b/TEMPLATE/_mapdata/init.sls
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
+---
+{#- Get the `tplroot` from `tpldir` #}
+{%- set tplroot = tpldir.split('/')[0] %}
+{%- from tplroot ~ "/map.jinja" import TEMPLATE as mapdata with context %}
+
+{%- do salt['log.debug']('### MAP.JINJA DUMP ###\n' ~ mapdata | yaml(False)) %}
+
+{%- set output_file = '/tmp/salt_mapdata_dump.yaml' %}
+
+{{ tplroot }}-mapdata-dump:
+  file.managed:
+    - name: {{ output_file }}
+    - source: salt://{{ tplroot }}/_mapdata/_mapdata.jinja
+    - template: jinja
+    - context:
+        map: {{ mapdata | yaml }}
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -156,6 +156,7 @@ suites:
       state_top:
         base:
           '*':
+            - TEMPLATE._mapdata
             - TEMPLATE
       pillars:
         top.sls:
@@ -177,6 +178,7 @@ suites:
       state_top:
         base:
           '*':
+            - TEMPLATE._mapdata
             - TEMPLATE
       pillars:
         top.sls:
--- a/test/integration/default/controls/_mapdata_spec.rb
+++ b/test/integration/default/controls/_mapdata_spec.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# Replace per minion strings
+replacement = {
+  hostname: system.hostname
+}
+
+# Keep only first 2 digits from Ubuntu finger
+mapdata_file = "_mapdata/#{system.platform[:finger].split('.').first}.yaml"
+
+# Load the mapdata from profile https://docs.chef.io/inspec/profiles/#profile-files
+mapdata_dump = inspec.profile.file(mapdata_file) % replacement
+
+control '`map.jinja` YAML dump' do
+  title 'should contain the lines'
+
+  describe file('/tmp/salt_mapdata_dump.yaml') do
+    it { should exist }
+    its('content') { should eq mapdata_dump }
+  end
+end
--- a/test/integration/default/files/_mapdata/amazonlinux-1.yaml
+++ b/test/integration/default/files/_mapdata/amazonlinux-1.yaml
@@ -1,0 +1,34 @@
+# yamllint disable rule:indentation rule:line-length
+# Amazon Linux AMI-2018
+---
+added_in_defaults: defaults_value
+added_in_lookup: lookup_value
+added_in_pillar: pillar_value
+arch: amd64
+config: /etc/template-formula.conf
+lookup:
+  added_in_lookup: lookup_value
+  master: template-master
+  winner: lookup
+master: template-master
+pkg:
+  name: cronie
+rootgroup: root
+service:
+  name: crond
+subcomponent:
+  config: /etc/TEMPLATE-subcomponent-formula.conf
+tofs:
+  files_switch:
+  - any/path/can/be/used/here
+  - id
+  - roles
+  - osfinger
+  - os
+  - os_family
+  source_files:
+    TEMPLATE-config-file-file-managed:
+    - example.tmpl.jinja
+    TEMPLATE-subcomponent-config-file-file-managed:
+    - subcomponent-example.tmpl.jinja
+winner: pillar
--- a/test/integration/default/files/_mapdata/centos-6.yaml
+++ b/test/integration/default/files/_mapdata/centos-6.yaml
@@ -1,0 +1,34 @@
+# yamllint disable rule:indentation rule:line-length
+# CentOS-6
+---
+added_in_defaults: defaults_value
+added_in_lookup: lookup_value
+added_in_pillar: pillar_value
+arch: amd64
+config: /etc/template-formula.conf
+lookup:
+  added_in_lookup: lookup_value
+  master: template-master
+  winner: lookup
+master: template-master
+pkg:
+  name: cronie
+rootgroup: root
+service:
+  name: crond
+subcomponent:
+  config: /etc/TEMPLATE-subcomponent-formula.conf
+tofs:
+  files_switch:
+  - any/path/can/be/used/here
+  - id
+  - roles
+  - osfinger
+  - os
+  - os_family
+  source_files:
+    TEMPLATE-config-file-file-managed:
+    - example.tmpl.jinja
+    TEMPLATE-subcomponent-config-file-file-managed:
+    - subcomponent-example.tmpl.jinja
+winner: pillar
--- a/test/integration/share/README.md
+++ b/test/integration/share/README.md
@@ -1,0 +1,20 @@
+# InSpec Profile: `share`
+
+This shows the implementation of the `share` InSpec [profile](https://github.com/inspec/inspec/blob/master/docs/profiles.md).
+
+Its goal is to share the libraries between all profiles.
+
+## Libraries
+
+### `system`
+
+The `system` library provides easy access to system dependent information:
+
+- `system.hostname`: return the result of `hostname -s` or `hostnamectl --static` based on the availability of either commands
+- `system.platform`: based on `inspec.platform`, modify to values that are more consistent from a SaltStack perspective
+  - `system.platform[:family]` provide a family name for Arch
+  - `system.platform[:name]` modify `amazon` to `amazonlinux`
+  - `system.platform[:release]` tweak Arch and Amazon Linux:
+    - `Arch` is always `base-latest`
+    - `Amazon Linux` release `2018` is resolved as `1`
+  - `system.platform[:finger]` is the concatenation of the name and the major release number (except for Ubuntu, which gives `ubuntu-20.04` for example)
--- a/test/integration/share/README.md
+++ b/test/integration/share/README.md
@@ -10,7 +10,7 @@ Its goal is to share the libraries between all profiles.
 
 The `system` library provides easy access to system dependent information:
 
-- `system.hostname`: return the result of `hostname -s` or `hostnamectl --static` based on the availability of either commands
+- `system.hostname`: return the result of `hostname -s` or `hostnamectl --static` based on the availability of either command
 - `system.platform`: based on `inspec.platform`, modify to values that are more consistent from a SaltStack perspective
   - `system.platform[:family]` provide a family name for Arch
   - `system.platform[:name]` modify `amazon` to `amazonlinux`
--- a/test/integration/share/inspec.yml
+++ b/test/integration/share/inspec.yml
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# vim: ft=yaml
+---
+name: share
+title: InSpec shared resources
+maintainer: SaltStack Formulas
+license: Apache-2.0
+summary: shared resources
+supports:
+  - platform-name: debian
+  - platform-name: ubuntu
+  - platform-name: centos
+  - platform-name: fedora
+  - platform-name: opensuse
+  - platform-name: suse
+  - platform-name: freebsd
+  - platform-name: amazon
+  - platform-name: arch
--- a/TEMPLATE/_mapdata/_mapdata.jinja
+++ b/TEMPLATE/_mapdata/_mapdata.jinja
@@ -1,0 +1,13 @@
+# yamllint disable rule:indentation rule:line-length
+# {{ grains.get('osfinger', grains.os) }}
+---
+{#- use salt.slsutil.serialize to avoid encoding errors on some platforms #}
+{{ salt['slsutil.serialize'](
+     'yaml',
+     map,
+     default_flow_style=False,
+     allow_unicode=True,
+   )
+   | regex_replace("^\s+'$", "'", multiline=True)
+   | trim
+}}
--- a/test/integration/default/controls/packages_spec.rb
+++ b/test/integration/default/controls/packages_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # Prepare platform "finger"
-platform_finger = "#{platform[:name]}-#{platform[:release].split('.')[0]}"
+platform_finger = system.platform[:finger].split('.').first.to_s
 
 control 'TEMPLATE package' do
   title 'should be installed'
@@ -9,7 +9,7 @@ control 'TEMPLATE package' do
   # Overide by `platform_finger`
   package_name =
     case platform_finger
-    when 'centos-6', 'amazon-2018'
+    when 'centos-6', 'amazonlinux-1'
       'cronie'
     else
       'bash'
--- a/test/integration/default/controls/services_spec.rb
+++ b/test/integration/default/controls/services_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # Prepare platform "finger"
-platform_finger = "#{platform[:name]}-#{platform[:release].split('.')[0]}"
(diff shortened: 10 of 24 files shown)


test/integration/default/controls/_mapdata_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Replace per minion strings
+replacement = {
+  hostname: system.hostname
+}
+
+mapdata_file = "_mapdata/#{system.platform[:finger].split('.').first}.yaml"
+mapdata_dump = inspec.profile.file(mapdata_file) % replacement
```
n-rodriguez, line 9: Maybe add a line to explain what's going here : what does `inspec.profile.file(mapdata_file)` return?

n-rodriguez (approved):

test/integration/default/controls/_mapdata_spec.rb
baby-gnu: This is an InSpec feature to [access files stored as resources in the current profile](https://docs.chef.io/inspec/profiles/#profile-files).

I can add a comment with the URL to the documentation.

daks: I wonder if `inputs` https://docs.chef.io/inspec/inputs/ could also be used simply (without us doing the job to find the correct file).

daks (approved):

test/integration/default/controls/_mapdata_spec.rb
baby-gnu: @daks I'm not sure it will make things easier :thinking:

n-rodriguez: > I can add a comment with the URL to the documentation.

yes please :+1:

daks: @baby-gnu not sure neither, was just an idea, I have no problem with the current implementation

n-rodriguez:
@baby-gnu can we merge?

myii:
> @baby-gnu can we merge?

@n-rodriguez Not yet, I've got a couple of minor changes that need to be made.  I'll just get around to them soon.

myii:
@saltstack-formulas/wg The first issue is that we need to finalise whether we're continuing with text-based or YAML-based comparisons.  The full comment is [here](https://github.com/saltstack-formulas/openssh-formula/pull/193#issuecomment-682430080); just quoting the table below:

> Factor | Text (current) | YAML (proposed)
> -- | -- | --
> InSpec diff output | `[+]` Line-by-line, much easier to see the differences | `[-]` Partial data structure, difficult to identify the differences
> Windows | `[-]` Hacky, have to resort to using `gsub` for `\n` => `\r\n` | `[+]` Works in the same way as on Linux, no workarounds required
> Comparison file ordering | `[?]` Must be exact (alphabetical throughout) | `[?]` As long as the data is correct, the ordering doesn’t matter

As I mentioned there, I'm favouring sticking with the text-based comparison, even though that's less "pure" -- it simply easier to work with from a human perspective.

The decision here will also affect [this PR](https://github.com/saltstack-formulas/sudoers-formula/pull/69).

kitchen.yml
```diff
@@ -155,6 +155,7 @@ suites:
       state_top:
         base:
           '*':
+            - TEMPLATE._mapdata
```
myii, line 158: Will need this line below for the `centos6` suite as well.

---

In fact, the includes/excludes need to be fixed here as well, which I'll probably do in a separate PR first.  So with that fix but without this line below, we have it failing:

* https://travis-ci.org/github/myii/template-formula/builds/723150877

I'll get this working and then mention the further changes that are needed here.

---

And while I've been adding the rest of the comments here, this is the test run with the actual diffs:

* https://travis-ci.org/github/myii/template-formula/builds/723154962

TEMPLATE/_mapdata/init.sls
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# vim: ft=sls
+---
+{#- Get the `tplroot` from `tpldir` #}
+{%- set tplroot = tpldir.split('/')[0] %}
+{%- from tplroot ~ "/map.jinja" import TEMPLATE as mapdata with context %}
```
myii, line 6: (`ssf`: Need standard method that can be used across all formulas -- will describe in a subsequent comment).

test/integration/default/controls/_mapdata_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# Replace per minion strings
+replacement = {
+  hostname: system.hostname
+}
```
myii, line 6: Should this be removed since it's not actually being used in this formula?

test/integration/default/controls/_mapdata_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+# Keep only first 2 digits from Ubuntu finger
+mapdata_file = "_mapdata/#{system.platform[:finger].split('.').first}.yaml"
+
+# Load the mapdata from profile https://docs.chef.io/inspec/profiles/#profile-files
+mapdata_dump = inspec.profile.file(mapdata_file) % replacement
```
myii, line 12: If removing the section above, then this becomes:

```suggestion
mapdata_dump = inspec.profile.file(mapdata_file)
```

myii, line 12: Just starting a conversation here but shouldn't we keep these in the narrowest possible scope, i.e. moving these into the `control` scope?

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
+# InSpec Profile: `share`
```
myii, line 1: This file differs substantially with `test/integration/share/README.md` in other formulas (which are already managed by the `ssf-formula`).  However, the changes here are very useful, so we'd need to update the template to include these.  One formula where this would get a little more confusing is the [libvirt-formula](https://github.com/saltstack-formulas/libvirt-formula/blob/6f1400e0021f16765f774aaecf58eda03fbebb4a/test/integration/share/README.md), where there are multiple libraries.  It would be extremely helpful if we could standardise these so that the template can be updated accordingly.

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+This shows the implementation of the `share` InSpec [profile](https://github.com/inspec/inspec/blob/master/docs/profiles.md).
+
+It's goal is to share the libraries between all profiles.
```
myii, line 5:
```suggestion
Its goal is to share the libraries between all profiles.
```

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Libraries
+
+### `system`
+
+The `system` library provides easy access to system dependents informations:
```
myii, line 11:
```suggestion
The `system` library provides easy access to system dependent information:
```

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+- `system.hostname`: return the result of `hostname -s` or `hostnamectl --static` based on the availability of each commans
```
myii, line 13:
```suggestion
- `system.hostname`: return the result of `hostname -s` or `hostnamectl --static` based on the availability of either command
```

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+- `system.platform`: take `inspec.platform` and mangle things a bit to be useful
```
myii, line 14:
```suggestion
- `system.platform`: based on `inspec.platform`, modify to values that are more consistent from a SaltStack perspective
```

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+  - `system.platform[:family]` provides a family name for Arch
```
myii, line 15:
```suggestion
  - `system.platform[:family]` provide a family name for Arch
```

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+  - `system.platform[:name]` modify `amazon` to `amazonlinux`
+  - `system.platform[:release]` tweak for Arch and Amazon Linux:
```
myii, line 17:
```suggestion
  - `system.platform[:release]` tweak Arch and Amazon Linux:
```

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+    - `Arch` is always `base-later`
```
myii, line 18:
```suggestion
    - `Arch` is always `base-latest`
```

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+    - `Amazon Linux` release `2018` became `1`
```
myii, line 19:
```suggestion
    - `Amazon Linux` release `2018` is resolved as `1`
```

test/integration/share/README.md
```diff
@@ -0,0 +1,20 @@
...
+  - `system.platform[:finger]` is just the concatenation of the name and the first release number (except for Ubuntu which gives `20.04` for example)
```
myii, line 20:
```suggestion
  - `system.platform[:finger]` is just the concatenation of the name and the first release number (except for Ubuntu which gives `ubuntu-20` for example)
```

test/integration/default/files/_mapdata/amazonlinux-1.yaml
```diff
@@ -0,0 +1,34 @@
+# yamllint disable rule:indentation rule:line-length
```
myii, line 1: This file will probably need updating as I fix the `centos6` suite (which really should be renamed as well).

test/integration/default/files/_mapdata/centos-6.yaml
```diff
@@ -0,0 +1,34 @@
+# yamllint disable rule:indentation rule:line-length
```
myii, line 1: This file will probably need updating as I fix the `centos6` suite (which really should be renamed as well).

test/integration/default/controls/_mapdata_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+control '`map.jinja` YAML dump' do
+  title 'should contain the lines'
+
+  describe file('/tmp/salt_mapdata_dump.yaml') do
+    it { should exist }
```
myii, line 18: These two lines depend on the conclusion of https://github.com/saltstack-formulas/template-formula/pull/203#issuecomment-684942520 -- i.e. could be `describe yaml` instead.

myii (changes requested):
While it looks like a lot of comments, most of these are trivial issues -- thanks for the PR!

---

* Checked using https://github.com/myii/ssf-formula/pull/249

kitchen.yml
myii: OK, I've fixed it:

* https://travis-ci.org/github/myii/template-formula/builds/723159868

However, there were a lot of minor changes required, that aren't relevant to this PR.  Let me submit a PR to fix those first and then this PR can be rebased on top.

test/integration/default/files/_mapdata/amazonlinux-1.yaml
```diff
@@ -0,0 +1,34 @@
...
+lookup:
+  added_in_lookup: lookup_value
+  master: template-master
+  winner: lookup
+master: template-master
+pkg:
+  name: bash
+rootgroup: root
+service:
+  name: systemd-journald
```
myii, line 18: Once the [new PR](https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481297085) is merged, this will become:

```suggestion
pkg:
  name: cronie
rootgroup: root
service:
  name: crond
```

test/integration/default/files/_mapdata/centos-6.yaml
```diff
@@ -0,0 +1,34 @@
...
+lookup:
+  added_in_lookup: lookup_value
+  master: template-master
+  winner: lookup
+master: template-master
+pkg:
+  name: bash
+rootgroup: root
+service:
+  name: systemd-journald
```
myii, line 18: Once the [new PR](https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481297085) is merged, this will become:

```suggestion
pkg:
  name: cronie
rootgroup: root
service:
  name: crond
```

test/integration/share/README.md
myii: Actually, this isn't only `ubuntu-20` is it?  It will resolve as `ubuntu-20.04`, right?  If that's the case, then perhaps this rewording is better:

```suggestion
  - `system.platform[:finger]` is the concatenation of the name and the major release number (except for Ubuntu, which gives `ubuntu-20.04` for example)
```

test/integration/default/files/_mapdata/amazonlinux-1.yaml
myii: Proposed fix: https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481298635.

test/integration/default/files/_mapdata/centos-6.yaml
myii: Proposed fix: https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481298806.

test/integration/share/inspec.yml
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# vim: ft=yaml
+---
+name: shared
+title: InSpec shared resources
```
myii, line 5: Similar to [this situation](https://github.com/saltstack-formulas/template-formula/pull/203/files#r481273581), this differs from all of the other `test/integration/share/inspec.yml` files out there.  The current template used by the `ssf-formula` produces:

```suggestion
name: share
title: TEMPLATE formula
```

The `name` must be changed (to match the actual suite name itself) but we need to decide what to do about the `title`, since it will affect all of the other formulas as well.

kitchen.yml
myii: Fix provided: #204.

test/integration/default/files/_mapdata/amazonlinux-1.yaml
myii: Fix provided: #204.

test/integration/default/files/_mapdata/centos-6.yaml
myii: Fix provided: #204.

test/integration/default/files/_mapdata/centos-6.yaml
myii: Fix provided: #204.

myii:
@baby-gnu Fixed underlying issues in #204.  If all is OK there and it is merged, you could also update this PR to use your new library to replace the old style `platform_finger` calculation that I've added in both test files there.

test/integration/default/controls/_mapdata_spec.rb
baby-gnu: It's not used and does nothing. I was wondering if we should let this as an example.

TEMPLATE/_mapdata/init.sls
baby-gnu: Sure, we need a way to find out which version of `map.jinja` we are using.

test/integration/default/controls/_mapdata_spec.rb
baby-gnu: Ho yes, you did that for `openntpd`. I don't know if there is an InSpec best practices for this but I personaly try to avoid doing much logic in the `control/end` block like I avoid the same in SaltStack states block ;-)

test/integration/default/controls/_mapdata_spec.rb
baby-gnu: Yes, I used `eq` for now since it's what is actually used for [openssh-formula](https://github.com/saltstack-formulas/openssh-formula).

It will be changed according to the result of the discussion.

baby-gnu:
> @baby-gnu Fixed underlying issues in #204. If all is OK there and it is merged, you could also update this PR to use your new library to replace the old style `platform_finger` calculation that I've added in both test files there.

Yes, I think it will be done in a dedicated PR.

myii:
A quick summary of where we are right now:

- [ ] [1] [Use text-based instead of YAML-based comparison?](https://github.com/saltstack-formulas/template-formula/pull/203#issuecomment-684942520) including https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481285583
  - _No change required (at least for the time being)_: Just use text-based comparison and update the PRs in other formulas to do the same.

- [ ] [2] [Set up `mapdata` in `_mapdata/init.sls`](https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481257608)
  - _No change required (at least for the time being)_: Merge as-is and resolve this when centralising the solution in the `ssf-formula`.

- [ ] [3] [Remove `replacement` from `_mapdata_spec.rb`?](https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481259338) including https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481259966
  - _No change required (at least for the time being)_: I don't believe there will be many formulas that will need this so no need to add the extra layer of complexity; however, can be resolved when centralising the solution in the `ssf-formula`.

- [ ] [4] [Use `control` scope for variables in InSpec?](https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481262937)
  - _No change required (at least for the time being)_: Merge as-is and open up for further discussion; @n-rodriguez, do you know anything about this?

- [ ] [5] [Change `commands` => `command`](https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481275489)
  - @baby-gnu, it would be better to get this fixed before merging; perhaps I can add a tiny commit to do this.

- [ ] [6] [Use `InSpec shared resources` across all formulas?](https://github.com/saltstack-formulas/template-formula/pull/203#discussion_r481306679)
  - _No change required (at least for the time being)_: Merge as-is and resolve this when centralising the solution in the `ssf-formula`.

- [ ] [7] [Replace old `platform_finger` concatenation with shared library method instead](https://github.com/saltstack-formulas/template-formula/pull/203#issuecomment-685035321)
  - _No change required (at least for the time being)_: Either resolve in an extra commit or leave for a subsequent PR.

Essentially, we can defer a lot of these for later.  Just need to deal with number 5 and perhaps number 7.

myii (approved):
Resolved numbers 5 and 7 with d68ed45109aa1274c6bf236db30758d795a3ba2a.

myii:
@baby-gnu @n-rodriguez @daks Merged, thanks for your contributions to this PR!

PR #238, opened 2021-08-18
test(pillar): use static test/salt/pillar/top.sls

baby-gnu:
The kitchen-salt provisionner have the `pillars_from_directories` option to recusively copy directories under target pillar root.

This has 3 advantages:

- simplify `kitchen.yml`

- manage pillar assignment with standard salt targetting mechanism, this avoid the dedicated `gentoo` suite (it could have been done from `kitchen.yml` itself by the way)

- ease the test outside kitchen by running `salt-call` directly with `--pillar-root` like:
  ```
  salt-call --local --id test-minion.example.net --file-root=template-formula/ --pillar-root=template-formula/test/salt/pillar/ state.show_sls TEMPLATE
  ```

* test/salt/pillar/top.sls: limit `gentoo` pillars based on `os_family` grain.

* kitchen.yml (suites): remove the now useless `gentoo` suite.
  Define an empty required `pillars` to not override static `top.sls`
  Define `pillars_from_directories` to copy them under `pillar_root`.

- [ ] Reviews completed

---

#### Primary type

- [ ] `[fix]`      A bug fix

#### Secondary type

- [x] `[test]`     Adding missing or correcting existing tests

No.

### Documentation checklist

### Testing checklist


myii:
Thanks, @baby-gnu.  This looks like a nice idea.

This will also necessitate some changes in how the `ssf-formula` works, so I'm going to need a little while to get around to this.  Currently got a bit of a backlog due to the `3003.2` release, etc.

baby-gnu:
> This will also necessitate some changes in how the `ssf-formula` works, so I'm going to need a little while to get around to this. Currently got a bit of a backlog due to the `3003.2` release, etc.

Thanks, it was an idea inspired by a user in SaltStack matrix room, it's not urgent.

kitchen.yml
```diff
@@ -429,64 +429,17 @@ verifier:
...
-            '*':
-              - TEMPLATE
-              - gentoo
-              - define_roles
-      pillars_from_files:
-        TEMPLATE.sls: pillar.example
-        gentoo.sls: test/salt/pillar/gentoo.sls
-        define_roles.sls: test/salt/pillar/define_roles.sls
+      pillars_from_directories:
+        - test/salt/pillar
```
myii, line 442: Will this still find `pillar.example` for the `TEMPLATE` pillar if it only looks in `test/salt/pillar`?

Or will we need to use `pillars_from_files` alongside `pillars_from_directories`?

```suggestion
      pillars_from_files:
        TEMPLATE.sls: pillar.example
      pillars_from_directories:
        - test/salt/pillar
```

kitchen.yml
```diff
@@ -429,64 +429,17 @@ verifier:
...
-          '*':
-            - TEMPLATE._mapdata
-            - TEMPLATE
-      pillars:
-        top.sls:
-          base:
-            '*':
-              - TEMPLATE
-              - gentoo
-              - define_roles
```
myii, line 485: We probably can avoid a dedicated `test/salt/pillar/top.sls` and just adapt the same method used in other formulas under `state_top`.

Some examples:

* [logrotate-formula](https://github.com/saltstack-formulas/logrotate-formula/blob/be7d5795ecd625a9e0427c1b81f328144b1c78d7/kitchen.yml#L453-L457)
* [ntp-formula](https://github.com/saltstack-formulas/ntp-formula/blob/878a101f6cedb88885798d9c42d157a11593fcbe/kitchen.yml#L453-L457)
* [rng-tools-formula](https://github.com/saltstack-formulas/rng-tools-formula/blob/0a0d041bcad00669244999c52a4466401a97ee01/kitchen.yml#L453-L457)

```suggestion
      pillars:
        top.sls:
          base:
            '*':
              - TEMPLATE
              - define_roles
            'G@os_family:Gentoo':
                - gentoo
```

test/salt/pillar/top.sls
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# vim: ft=yaml
+---
+base:
+  '*':
+    - TEMPLATE
+    - define_roles
+  'os_family:Gentoo':
+    - match: grain
+    - gentoo
```
myii, line 10: While this doesn't affect this specific PR, there's another slight issue to consider here.

The ordering of the entries is significant, since the pillars are merged in the order given.  So we'll now be merging the `gentoo` pillar at the very end, rather than between `TEMPLATE` and `define_roles`.  What will we do about situations where we need the platform-specific pillar merged in the middle?

Of course, this applies whether we keep this file or keep this in `kitchen.yml` as mentioned in the previous comment.

myii (commented):
@baby-gnu I like the idea of this PR and I reckon we could benefit from using `pillars_from_directories` in various formulas.  However, we've got a few situations to consider, which I've mentioned inline.

NOTE: I haven't had a chance to actually test any of the ideas suggested, yet.

kitchen.yml
baby-gnu: You are completely right, in my PR, I [kept](https://github.com/saltstack-formulas/template-formula/blob/7fc998282d9ee679db6a6ccb1cb76b2bf79938eb/kitchen.yml#L438) the `TEMPLATE.sls: pillar.example` in `pillars_from_files`

kitchen.yml
baby-gnu: Yes, It could be done from `kitchen.yml` directly but the main point of having a dedicated `top.sls` is to be able to run it outside kitchen.

test/salt/pillar/top.sls
baby-gnu: In some situation I need to apply some pillar at the very last for all minions, I do something like the following:

```yaml
# -*- coding: utf-8 -*-
# vim: ft=yaml
---
base:
  '*':
    - TEMPLATE
  'os_family:Gentoo':
    - match: grain
    - gentoo
  '*':
    - define_roles
...
```

kitchen.yml
myii: @baby-gnu Ah, the diff caught me out, with the two suites being merged into one.

kitchen.yml
myii: @baby-gnu I haven't come across this idea before -- could you give me an example of this being run outside of Kitchen?  Does this also apply to the using the `state_top`'s `top.sls` as a separate file as well?

test/salt/pillar/top.sls
myii: @baby-gnu So that works?  That's actually a `yamllint` violation ([`key-duplicates`](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.key_duplicates)).  I suppose we could use a `yamllint` ignore for those lines, if we ever need them.

kitchen.yml
baby-gnu: The idea came [from user colttt on SaltStack matrix room](https://matrix.to/#/!WmgUPuThcUZvfCoQLZ:matrix.org/$162920593550701UaMAC:matrix.org) to better validate formula since `salt-lint` does not cover all possible errors.

colttt uses the following command to compile the states:

```
salt-call --id=test-minion.example.de --local --file-root=/srv/salt/states --pillar-root=/srv/salt/pillars --retcode-passthrough state.show_sls
```

But it can't be done easily when the formula requires pillars.

Then, my idea is to provide all the necessary to try out a formula in two steps:

1. clone the formula:
   ```
   git clone https://github.com/saltstack-formulas/template-formula.git && cd template-formula

   ```
2. execute the formula locally with our test pillar values:
   ```
   salt-call --local --id test-minion.example.net --file-root=./ --pillar-root=./test/salt/pillar/ state.show_sls TEMPLATE
   ```

There is no point in having `state_top`'s `top.sls` in a dedicated file too since there is no `salt-call` option to select that file and I think we should avoid having one in the formula directory referenced by the `--file-root` option above.

test/salt/pillar/top.sls
baby-gnu: Ho yes, my bad, I thought I did that but in fact I cheated with:

```yaml
base:
  '*':
    - global_vars
  'domain:example.net':
    - match: grain
    - example_net
# […]
  'os:*':
    - match: grain
    - mine
```

kitchen.yml
myii: @baby-gnu The example you've shared doesn't appear to work, because the `TEMPLATE` pillar (i.e. `pillar.example`) is not in that directory:

```sls
[ERROR   ] Specified SLS 'TEMPLATE' in environment 'base' is not available on the salt master
[ERROR   ] Template was specified incorrectly: False
[CRITICAL] Pillar render error: Specified SLS 'TEMPLATE' in environment 'base' is not available on the salt master
Error running 'state.show_sls': Pillar failed to render. Additional info follows:

- Specified SLS 'TEMPLATE' in environment 'base' is not available on the salt master
```

If I comment out the `TEMPLATE` line in the newly provided `top.sls`, then the state runs -- but obviously missing everything from `pillar.example` (the `TEMPLATE` pillar).

baby-gnu: Yes, that's an issue.

To be completely functional, the `test/salt/pillar/` should contains all the required files, in which case:
- the `pillars_from_files` is no more used
- the `pillar.example` is just an example and not a seed for tests.

baby-gnu: I push a second commit to move testing values from `pillar.example` to `test/salt/pillar/TEMPLATE.sls` and remove `pillars_from_files` from `kitchen.yml`.

myii: @baby-gnu If we're going to attempt to use these ideas across existing formulas then that's not going to be fun to refactor.  There are more formulas using `pillar.example` in the CI than any other file.  Due to that, it's the default value used from the `ssf-formula` as well.

* [50+ formulas](https://github.com/search?o=desc&q=org%3Asaltstack-formulas+filename%3Akitchen.yml+pillar.example&s=indexed&type=Code) across the whole org.
* Around [37 formulas](https://github.com/search?q=org%3Asaltstack-formulas+filename%3Akitchen.yml+%22pillar.example%22+%22pillars_from_files%22&type=Code) that are managed by the `ssf-formula`.

We can discuss the relative merits of this idea in the upcoming meeting.  I'm still not quite sure what the use cases are so it would great if you could demonstrate it.

myii:
@baby-gnu I thought I'd familiarise myself with these changes by adapting it for the `salt-formula`.  This is the commit I used, which is passing in the CI:

* https://github.com/myii/salt-formula/commit/29c955e906e4edf9ecbc50cd35170c17f28c6b43

A couple of things I came across so far, which are valid for this PR as well:

1. There's no need to use `pillars: {}` since [that's the default](https://github.com/saltstack/kitchen-salt/blame/2296170ad5642fb804030d65acc769dd9a9fb360/docs/provisioner_options.md#L247-L249) provided by `kitchen-salt`.
2. (I haven't done this in the commit there but) We probably need to remove the following line as well, since it prevents the `top.sls` from being `git` added expect using `-f`:

https://github.com/saltstack-formulas/template-formula/blob/8d506ea93eca196c60e178f8dd2eb2a7ad6594e9/.gitignore#L129

baby-gnu:
> * There's no need to use `pillars: {}` since [that's the default](https://github.com/saltstack/kitchen-salt/blame/2296170ad5642fb804030d65acc769dd9a9fb360/docs/provisioner_options.md#L247-L249) provided by `kitchen-salt`.

You are right, I'll remove it on the final rework of this PR

> * (I haven't done this in the commit there but) We probably need to remove the following line as well, since it prevents the `top.sls` from being `git` added expect using `-f`:

I'm really not sure, I think we should keep it to avoid having a `top.sls` in the states tree, so I propose to use:

```
top.sls
!test/salt/pillar/top.sls
```

This way, all `top.sls` will be ignored except the `test/salt/pillar/top.sls` one.

test/salt/pillar/top.sls
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# vim: ft=yaml
+---
+base:
+  '*':
+    - defaults
+    - define_roles
```
myii, line 7:
```suggestion
```

We've already got `define_roles` below using your "special hack".

Or perhaps we shouldn't confuse people who are using this formula as a template... and delete the two lines below instead.  What do you think?

myii (changes requested):

test/salt/pillar/top.sls
baby-gnu: My fault, it's a typo.

myii:
Thanks for the patience, @baby-gnu -- merged.

Rebased your changes on the latest version of this formula (the CI instances had changed).

The main notes regarding the additional commit:

1. Moved `pillars_from_directories` under `provisioner` instead of under `suites`, as discussed in the working group meetings.
2. Used `test/salt/pillar/default.sls` instead of `test/salt/pillar/defaults.sls`, to maintain consistency with the large number of formulas out there.
3. Other minor consistency fixes, such as updating `.travis.yml` and `CODEOWNERS`.

PR #252, opened 2022-11-01
refactor: moving jinja to separate dir for cleaner layout

vveliev-tc:

- [ ] Reviews completed

---

#### Primary type

- [ ] `[fix]`      A bug fix

#### Secondary type

No.

Simply moving jinja files to separate folder to make bit more cleaner root folder. (jinja may scare new contributors ;) )

### Documentation checklist

### Testing checklist


daks:
Looks fine to me, but I would prefer a more agnostic name like `libs` which could be used for other type of libraries if needed later.

Also, we must be sure it has no negative impact of unusual setups (like `salt-ssh` for example). ping @baby-gnu @myii

daks (commented):
Looks fine to me, but I would prefer a more 'agnostic' name like `libs` which could be used for other type of libraries if needed later.

Also, we must be sure it has no negative impact on unusual setups (like `salt-ssh` for example). ping @baby-gnu @myii

vveliev-tc:
I was actually not sure about the name for the folder, since there is a git ignore for `lib` https://github.com/saltstack-formulas/template-formula/blob/master/.gitignore#L18

baby-gnu:
> I was actually not sure about the name for the folder, since there is a git ignore for `lib` https://github.com/saltstack-formulas/template-formula/blob/master/.gitignore#L18
> 
> should it be `libs/jinja/map.jinja` or just `libs/map.jinja` ?

I personally prefer `libs/map.jinja`.

Regards.

daks:
If we all three agree for `/libs` directory, let's go for it! :)

dafyddj:
This might be a minor detail, but I would go for `lib` myself. The fact that `lib` is in `.gitignore` seems to be by accident rather than by design.

vveliev-tc:
let agree on the style before I change it again :).

daks (approved):
Thanks :)

baby-gnu (approved):

baby-gnu:
> Thanks :)

Now, I'm wondering how to translate this to `ssf-formula` :sweat_smile:

baby-gnu:
Hello, looking at this for another formula, I wonder if the per formula `post-map.jinja` shouldn't be under `parameters/` instead of `libs/`.

This `post-map.jinja` is more about specific data manipulations instead of a generic library.

As a user of a formula, I'll give more attention to files under `parameters/` to understand a formula rather than under `libs/`.

Only 2 formulas use this `post-map.jinja` for now:

- [consul-formula](https://github.com/saltstack-formulas/consul-formula/blob/2968032/consul/post-map.jinja)
- [java-formula](https://github.com/saltstack-formulas/java-formula/blob/3b8795e/java/post-map.jinja)

vveliev-tc:
> Hello, looking at this for another formula, I wonder if the per formula `post-map.jinja` shouldn't be under `parameters/` instead of `libs/`.
> 
> This `post-map.jinja` is more about specific data manipulations instead of a generic library.
> 
> As a user of a formula, I'll give more attention to files under `parameters/` to understand a formula rather than under `libs/`.
> 
> Only 2 formulas use this `post-map.jinja` for now:
> 
> * [consul-formula](https://github.com/saltstack-formulas/consul-formula/blob/2968032/consul/post-map.jinja?rgh-link-date=2022-12-22T10%3A56%3A39Z)
> * [java-formula](https://github.com/saltstack-formulas/java-formula/blob/3b8795e/java/post-map.jinja?rgh-link-date=2022-12-22T10%3A56%3A39Z)

I'm not sure to be honest sounds like `parameters/` will be better place for `post-map.jinja` as jinja files will be expected under parameters as well. 

does it need to be resolved for this PR?

baby-gnu:
> I'm not sure to be honest sounds like `parameters/` will be better place for `post-map.jinja` as jinja files will be expected under parameters as well.

I'm not sure to understand your point 🤔

To me:

- `libs/` is a good place for _code_ reusable between formulas, which could easily be managed by automation tools like [ssf-formula](https://github.com/myii/ssf-formula)
- `post-map.jinja` is not something generic but much more formula dependent and, to me, it makes sense to put it somewhere else

> does it need to be resolved for this PR?

This PR will be a starting point for new standard accross formulas, I think we should take time to make it clean and permit to everybody to give their points.

Regards.

vveliev-tc:
@baby-gnu  I think we are on the same page, I have updated map.jinja to reflect that. 

Is there anythinb else I should change?

TEMPLATE/libs/map.jinja
```diff
@@ -63,4 +63,4 @@
 
 {#- Per formula post-processing of `mapdata` if it exists #}
 {%- do salt["log.debug"]("map.jinja: post-processing of 'mapdata'") %}
-{%- include tplroot ~ "/post-map.jinja" ignore missing %}
+{%- include tplroot ~ "/parameters/post-map.jinja" ignore missing %}
```
baby-gnu, line 66: You can use `formula_param_dir` variable defined line 9.

baby-gnu (approved):
This is a good PR, my proposal of change is completely optional since it's just me being very picky.

TEMPLATE/libs/map.jinja
vveliev-tc: oh thanks, I totaly missed it!

vveliev-tc:
> Hey @vveliev-tc we noticed that you have been working on this again and we would like to figure out how we can work to get this moving forward. We aren't quite ready to get this merged in as @dafyddj is working on the copier/renovate work that would deploy this to the formulas managed with that methodology.
> 
> One discussion point that came up in the salt stack formulas working group call 2 weeks ago was the idea of using `_lib` or `_libs` as a way to avoid the existing conflict in formulas such as [hadoop-formula](https://github.com/saltstack-formulas/hadoop-formula/tree/master/hadoop/libs). This would also help to communicate that these are more private internal libraries used by the formulas.
> 
> I don't have a strong preference on if `_lib` or `_libs` is used but in some quick looking it seems libs is more often associated with external 3rd party dependencies and that has me leaning towards `_lib`
> 
> What are your thoughts and are there any ways we can help support your work with this PR?

Hey @jeff350 I just fixed github actions to see if PR can be merged as it already was approved.  

I there anything else I should change?

